PR #28, opened 2018-08-20
Add data size and modify KWs to register file

alanking:
Adds data size and modify keywords to the register_file operation to properly apply metadata from the filesystem without needing the extra call to dataObjModMeta after registering.

---
Note: This requires changes in iRODS server to be effective (see irods/irods#4066).


trel:
This change will raise the iRODS requirement to 4.2.4+.

PR #58, opened 2018-09-17
Add S3 support

alanking:
Adds following command line options to allow for scanning S3 buckets:

- s3_endpoint_domain
- s3_region_name
- s3_keypair
- s3_proxy_url

s3_keypair must point to a file on the local machine with an S3 keypair.
s3_proxy_url is only needed if using a proxy to connect to the S3 endpoint.

S3 buckets are scanned via "chunking" of the retrieved objects such that each task is delegated a chunk of objects, rather than just one. The number of objects to process can be configured with the files_per_task option.

scandir is now the only option for scanning filesystems. As such, list_dir and scan_dir_list command line options have been removed.

Logic for grouping objects to process per task in S3 is now used for filesystem scanner as well. This allows for a significant performance boost to avoid overhead of fetching tasks off the queue for everyfile or object which needs processing.

---
Tests passed (no new tests added).

### DIFF
--- a/irods_capability_automated_ingest/flask_app.py
+++ b/irods_capability_automated_ingest/flask_app.py
@@ -30,8 +30,7 @@ def put(job_name, data):
     data2.setdefault("timeout", 3600)
     data2.setdefault("interval", None)
     data2.setdefault("profile", False)
-    data2.setdefault("list_dir", False)
-    data2.setdefault("scan_dir_list", False)
+    data2.setdefault("files_per_task", 50)
     data2["job_name"] = job_name
     data2["config"] = get_config()
 
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ setup(
     keywords='irods automated ingest landingzone filesystem',
     packages=find_packages(),
     install_requires=[
+        'minio',
         'flask',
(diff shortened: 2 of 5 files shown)


irods_capability_automated_ingest/flask_app.py
```diff
@@ -30,8 +30,7 @@ def put(job_name, data):
     data2.setdefault("timeout", 3600)
     data2.setdefault("interval", None)
     data2.setdefault("profile", False)
-    data2.setdefault("list_dir", False)
-    data2.setdefault("scan_dir_list", False)
+    data2.setdefault("files_per_task", 1)
```
trel: there is an argument to be made that using '1' here preserves the existing earlier behavior.

however, i think a stronger argument might be that performance is what users want by default, so we should pick a stronger default.

do we need to wait for more empirical data before updating the default?

setup.py
```diff
@@ -31,6 +31,7 @@
     keywords='irods automated ingest landingzone filesystem',
     packages=find_packages(),
     install_requires=[
+        'boto3',
```
trel: don't need boto3, need minio instead

Issue #51, opened 2018-09-10
event handler callbacks for collections were not invoked for EMPTY directory / sub-directory


masilamani:
- [x] fix PEPs for sync_dir
- [x] add tests for event_handler PEP methods
---
Event handler callbacks for collections works fine for the directory with files but not for EMPTY directory / sub-directory.

PR #61, opened 2018-09-18
[#51] Fix policy points for syncing dirs

alanking:
Partial revert of 1710668deebe3ab39ae13e4dcc182559543e0d27

sync_dir is once again a separate operation for which policy can be invoked. pre/post_coll_create and *_modify should work for both empty and non-empty directories.


trel:
tests pass?    should we have a new test or two that exercises the policy endpoints?

Issue #31, opened 2018-08-21
check for valid source directory and collection name, zone name


masilamani:
Currently,   if we give incorrect zone name or collection name by mistake, it runs for ever and it does not check and throws an error saying invalid source directory path, or invalid zone name,  or invalid collection name/path.

masilamani:
For example: Try with giving invalid zone as target collection

python -m irods_capability_automated_ingest.irods_sync start /tmp/ingest /INVALIDZone/home/rods/tmp/ingest

alanking:
I was able to reproduce the issue with an invalid zone name. The job will hang and needs to be stopped manually.

I'm not sure how you would give an invalid collection name (perhaps if you didn't have permission to write in the target collection?), but after trying a few things I was not able to reproduce for that scenario. Using an invalid source directory results in the job failing out and stopping with `FileNotFoundError`. This issue is fairly old, so these cases may have been resolved somewhere along the way.

PR #84, opened 2018-12-13
[#31] Handle invalid zone name in target coll

alanking:
Now raise an exception when the target collection contains an invalid zone name.

Tests pass.

---
It would be nice to be able to catch this in the task manager before putting it on the queue but that would require an iRODS connection on the machine which kicks off the sync job.

### DIFF
--- a/irods_capability_automated_ingest/test/test_irods_sync.py
+++ b/irods_capability_automated_ingest/test/test_irods_sync.py
@@ -2,11 +2,11 @@ import unittest
 import os
 import os.path
 import stat
+import subprocess
 import glob
 import base64
 from unittest import TestCase
 from redis import StrictRedis
-from subprocess import Popen
 from signal import SIGINT
 from os import makedirs, listdir, remove
 from shutil import rmtree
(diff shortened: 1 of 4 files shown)


trel:
Yes, a pre-check would save potentially millions of connections with bad information.

alanking:
I have moved the target collection checks into a function in `sync_irods.py` which is called from `sync_task.py` as a pre-check for the scan job.

alanking:
Fixed trailing whitespace and unused variable. `subprocess` concerns already exist, are limited to the test file, and imply a larger change to the way tests are created.

For posterity, an initial thought would be to create arguments within a given test class, shove them into `sys.argv`, and run the sync tool via python rather than via `subprocess`.

irods_capability_automated_ingest/test/test_irods_sync.py
```diff
@@ -574,6 +585,21 @@ def test_register_non_leaf_non_root_with_resc_name(self):
...
+    def test_register_to_invalid_zone(self):
+        self.do_run_invalid_target_collection_test(
+            '/invalidZone/home/rods',
+            'Invalid zone name in destination collection path')
+
+    def test_register_to_existing_zone_substring(self):
+        self.do_run_invalid_target_collection_test(
+            '/tempZ/home/rods',
+            'Invalid zone name in destination collection path')
+
```
trel: maybe add one more test where it's a superstring... '/tempZoneMore/home/rods''

alanking: Done

trel:
looks good - one day we'll quiet codacy regarding subprocess...

Issue #63, opened 2018-09-19
Create missing tests


alanking:
- [x] Non-utf8-encodable filenames for register_as_replica_sync (see https://github.com/irods/irods_capability_automated_ingest/issues/251)
- [x] Non-utf8-encodable filenames for put_sync
- [x] Non-utf8-encodable filenames for put_append
- [x] Chunking/files_per_task logic (see https://github.com/irods/irods_capability_automated_ingest/issues/252)
- [x] File name/directory exclusion (filename exclusion test added in c508e17111d8b2fc24a1235b59fed36a77bd6c76) (see https://github.com/irods/irods_capability_automated_ingest/issues/253 for `--exclude_directory_name`)
- [x] Special file types (symlinks, sockets, pipes, etc.) (see https://github.com/irods/irods_capability_automated_ingest/issues/204)
- [x] Symlinks to directories (https://github.com/irods/irods_capability_automated_ingest/issues/254)
- [x] `--profile` option (see https://github.com/irods/irods_capability_automated_ingest/issues/255)

Add more as needed

alanking:
For documentation purposes...

Evaluating missing tests can be improved via a tool like [coverage.py](https://coverage.readthedocs.io/en/v4.5.x/). Using coverage is easy enough, but I am currently running into issues with tracing code invoked via `subprocess`. There are [instructions for measuring sub-processes](https://coverage.readthedocs.io/en/coverage-4.2/subprocess.html).

alanking:
Instructions for getting Coverage running...

For now, need custom install of coverage with checkout at this commit: https://github.com/elyezer/coveragepy/commit/fa2a94dccb34b017035ccb819b037cab88ff492f
Provides monkeypatch for Billiard hook in coverage module
Post detailing process for change above... https://www.redhat.com/archives/pulp-dev/2016-August/msg00018.html
gist with instructions for running tests... https://gist.github.com/elyezer/2d8225843451364cbb19af08f4f52de3
Celery has stated that version 5 will be replacing Billiard with something else: https://github.com/celery/celery/issues/3422#issuecomment-358974274

Follow steps here to set up coverage for subprocess: https://coverage.readthedocs.io/en/v4.5.x/subprocess.html

Need to have COVERAGE_PROCESS_START set to config file path:
```
export COVERAGE_PROCESS_START=/path/to/.coveragerc
```
OR, in the test file...
```
[...]

PR #87, opened 2019-01-16
[#63] Add more UnicodeEncodeError tests

alanking:
Adds test for put_sync and put_append.

Also adds test for register as replica sync, but this does not appear to be working at the moment, so it is being skipped like the other register as replica sync tests.

---
Tests passed on docker instance on local machine except where noted.


irods_capability_automated_ingest/test/test_irods_sync.py
```diff
@@ -1069,35 +1074,101 @@ def test_register(self):
...
             self.assert_metadata_annotation(session)
             self.assert_data_object_size(session)
 
+    def test_put_append(self):
+        expected_physical_path = join(DEFAULT_RESC_VAULT_PATH, 'home', 'rods', os.path.basename(self.source_dir_path), self.unicode_error_filename)
+
+        self.run_scan_with_event_handler("irods_capability_automated_ingest.examples.append")
+
+        with open(self.bad_filepath, 'a') as f:
+            f.write('test_put_append')
```
trel: self.create_bad_file() here as well?

trel: glossed over the 'a'.   right.

Issue #88, opened 2019-04-01
Error when starting Celery workers


alanking:
Running Celery workers after installing automated ingest yields the following error:
```
$ celery -A irods_capability_automated_ingest.sync_task worker -l error -Q restart,path,file -c16
 
 -------------- celery@xxx v4.3.0 (rhubarb)
---- **** ----- 
--- * ***  * -- Linux-3.10.0-862.2.3.el7.x86_64-x86_64-with-centos-7.5.1804-Core 2019-04-01 15:34:38
-- * - **** --- 
- ** ---------- [config]
- ** ---------- .> app:         icai:0x7f52e1781a20
- ** ---------- .> transport:   redis://xxx:6379/0
- ** ---------- .> results:     disabled://
- *** --- * --- .> concurrency: 16 (prefork)
-- ******* ---- .> task events: OFF (enable -E to monitor tasks in this worker)
--- ***** ----- 
 -------------- [queues]
                .> file             exchange=file(direct) key=file
                .> path             exchange=path(direct) key=path
                .> restart          exchange=restart(direct) key=restart

[2019-04-01 15:34:38,954: CRITICAL/MainProcess] Unrecoverable error: VersionMismatch('Redis transport requires redis-py versions 3.2.0 or later. You have 2.10.6',)
...
kombu.exceptions.VersionMismatch: Redis transport requires redis-py versions 3.2.0 or later. You have 2.10.6
```
Celery version changed to 4.3.0 which updated Kombu which now requires a newer version of redis. The new version of redis is known to break automated ingest, so we need to limit the version of Celery to 4.2.x for now.

PR #89, opened 2019-04-02
[#88] Limit Celery version

alanking:
Also limits billiard version due to interdependencies

Test suite passes at the bench. Versions can be upgraded in a controlled manner from here\


trel:
seems good - ready for # if you are

Issue #90, opened 2019-04-26
honor CELERY_BROKER_URL


trel:
The ingest tool currently requires `--redis_host`, `--redis_port`, and `--redis_db` if any of them need to be non-default values... (when celery is talking to a redis server elsewhere).

The ingest tool should honor the environment variable `CELERY_BROKER_URL` of the form:
```
CELERY_BROKER_URL=redis://hostname:port/db
```

If any of the `--redis...` options are passed directly, they should take precedence over the environment variable.

PR #106, opened 2019-11-18
[#109] Update docker steps for Celery

alanking:
Provides two Dockerfiles:

1. Automated ingest as a python application image

2. Celery worker image based on ingest image

Two additional environment files are required in and
outside of the container to make it run properly:

1. irods_environment.json -- This is not generated,
even with equivalent set of environment variables.

2. icommands.env -- This serves as the authentication
for our iRODS client.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,9 +1,8 @@
 FROM python:3.5
 
-MAINTAINER Hao Xu version: 0.1.0
+RUN pip install irods-capability-automated-ingest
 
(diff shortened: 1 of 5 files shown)


alanking:
Unfortunately, this still requires building of the Docker image to include the environment file with the correct authentication information or, alternatively, the user will need to volume mount an existing irods_environment.json and set `IRODS_ENVIRONMENT_FILE` environment variable at runtime.

The issues around `CELERY_BROKER_URL` will go away with the resolution of #90.

alanking:
This is a little more "in progress" than I thought, so I will open another PR when closer to a complete solution to avoid spamming watchers.

PR #113, opened 2019-12-02
[#109] Update docker steps for Celery

alanking:
Provides two Dockerfiles:

1. Automated ingest as a python application image
2. Celery worker image based on ingest image

icommands.env and irods_envirionment.json files are required in and
outside of the container to make it run properly. This serves as the
connection info/authentication for our iRODS client.

Also provides a set of wrapper bash scripts for easy use of the
containers:

1. Scripts for listing, starting, stopping, and watching ingest jobs.
2. "Point-and-click" script: Minimizes typing by standing up Celery
   workers, scanning the target to the destination logical path using
   mainly default settings, and tearing everything down at the end.

NOTE: All of these scripts - especially point_and_click.sh - make a
series of assumptions about where they are running and what is
available to them at runtime. Please do not use these in production.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,54 @@
+# How to dockerize with manual config
+
+An iRODS server is assumed to be running on an accessible network.
+
+## Step 1: Build
+Build the ingest image:
+```
+$ docker build -t ingest .
+```
+
+Build the Celery worker image (uses ingest image as a base so as to be identical):
+```
+$ docker build -t ingest-workers -f Dockerfile.ingest_worker .
+```
+
+## Step 2: Start redis server
+
+Start redis server in one container:
+```
+$ docker run --rm --name some-redis -d redis:4.0.8
+```
+
+## Step 3: Start Celery workers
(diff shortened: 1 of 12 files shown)


docker/README.md
```diff
@@ -0,0 +1,54 @@
...
+IRODS_USER_NAME=rods
+IRODS_ZONE_NAME=tempZone
+IRODS_PASSWORD=rods
+```
+This will serve as the iRODS client authentication (rather than running `iinit` somewhere).
+Also need a valid irods_environment.json file.
+
+This line will start 4 Celery workers with read-only access to a directory on the host machine:
+```
+docker run --rm --name workers -e "CELERY_BROKER_URL=redis://redis:6379/0" --env-file icommands.env --link some-redis:redis --link irods-provider:icat.example.org -v /path/to/ingest:/mount/path:ro ingest-worker -c 4
```
trel: should be plural?  to match the name of the image?
`ingest-workers`

trel: nope, i take it back, the earlier line 13 should be singular, instead.

Issue #114, opened 2019-12-03
Ingest cannot start with as_user() returning a rodsuser account


jasoncoposky:
The framework tests for the existence of the root collection '/' which is not readable by a rodsuser account which then throws an excption:
https://github.com/irods/irods_capability_automated_ingest/blob/master/irods_capability_automated_ingest/sync_irods.py#L23

trel:
@alanking Can we close this?  And therefore #138 too?

alanking:
I don't remember why we left this open, so I think it can be closed.

I think the original reasoning for the root collection check was a false premise anyway.

trel:
Okay, closing and marking the other as duplicate.    We'll make a new issue if it comes up again.

PR #115, opened 2019-12-04
[#114] Remove zone hint check

alanking:
Zone hint and root collection cannot be seen by rodsusers, so running
ingest as any user other than a rodsadmin currently does not work. The
checks have been removed for now. This is a partial revert of SHA:
ecf5d1e95b9ae41de31c1a6ecd9ecb0bd36b0e3d

Also add NetworkException to sync_irods.


trel:
looks good - please to #

alanking:
Opened #116 to address the hanging jobs situation.

PR #128, opened 2020-03-02
fixed the parsing of the S3 region parameter

ccacciari:
Apparently the Minio client was instantiated without the S3 region parameter. I added it to the the list of input parameters.


trel:
Thanks!

PR #134, opened 2020-07-01
put_sync functionality for data in S3

acnewton:
Added the functionality to transfer the data from an object in S3 to an object in iRODS. Not only registering the file in place. Using Minio library to do the transfer. It is also possible to append to a file by setting a certain offset. The md5sum hash is calculated during streaming. Final calculated hash is compared to the Etag header from the S3 object. Note that this Etag is not always the md5sum, when the file has been uploaded via multipart upload. A more general way of comparing checksum will be necessary.


alanking:
Thanks for this! We will put it through once we have a chance to get some testing built up around it. Sorry it has taken so long to respond

alanking:
Superseded by #206, wherein this work has been preserved. Thanks!

Issue #157, opened 2021-09-09
$IRODS_ENVIRONMENT_FILE required to run ingest tool


alanking:
There is no point in considering the environment variables `$IRODS_HOST`, `$IRODS_PORT`, `$IRODS_USER_NAME`, `$IRODS_ZONE_NAME`, `$IRODS_PASSWORD` if `$IRODS_ENVIRONMENT_FILE` is also required to use the ingest tool as is the case today. The environment variables should be sufficient for the worker nodes to run.

alanking:
Duplicate of #147, whoops

Issue #147, opened 2021-01-12
os.environ entries cannot be None


holtgrewe:
https://github.com/irods/irods_capability_automated_ingest/blob/4207df77bd763838aab57f08081eda8900bcb400/irods_capability_automated_ingest/irods_sync.py#L31

Rather use `if not os.environ.get("<key>"):` and rely on pythonic truthyness of `""` and `None`.

PR #158, opened 2021-09-09
[#147][#157] Allow running workers with env only

alanking:
`$IRODS_ENVIRONMENT_FILE` is still supported but if the env contains the
supported environment variables, this makes deployment of worker nodes
much easier. Here are the supported/expected environment variables:
```
    IRODS_HOST
    IRODS_PORT
    IRODS_USER_NAME
    IRODS_ZONE_NAME
    IRODS_PASSWORD
```
If any of these is missing, an environment file in JSON format is
expected to be at the path specified by `$IRODS_ENVIRONMENT_FILE`. The
environment file should contain at least the following information:
```
    {
        "irods_host": <string>,
        "irods_port": <int>,
        "irods_user": <string>,
        "irods_zone_name": <string>
    }
```
An `.irodsA` file is expected to be in the home directory of the Linux
user running the Celery workers on a given node in order to authenticate
with iRODS.

### DIFF
--- a/irods_capability_automated_ingest/sync_irods.py
+++ b/irods_capability_automated_ingest/sync_irods.py
@@ -310,12 +310,15 @@ def irods_session(handler_module, meta, logger, **options):
 
     key = json.dumps(kwargs) # todo add timestamp of env file to key
 
-    with open(env_file) as irods_env:
-        irods_env_as_json =  json.load(irods_env)
-        if 'irods_ssl_ca_certificate_file' in irods_env_as_json:
-            kwargs['ssl_context'] = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH,
-                                                               cafile=irods_env_as_json['irods_ssl_ca_certificate_file'],
-                                                               capath=None, cadata=None)
+    if env_file:
+        with open(env_file) as irods_env:
+            irods_env_as_json =  json.load(irods_env)
+            verify_server = irods_env_as_json.get('irods_ssl_verify_server')
+            ca_file = irods_env_as_json.get('irods_ssl_ca_certificate_file')
+            if verify_server and verify_server != 'none' and ca_file:
+                kwargs['ssl_context'] = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH,
+                                                                   cafile=ca_file,
+                                                                   capath=None, cadata=None)
 
     if not key in irods_session_map:
         # TODO: #42 - pull out 10 into configuration


irods_capability_automated_ingest/sync_irods.py
```diff
@@ -310,12 +310,13 @@ def irods_session(handler_module, meta, logger, **options):
...
-    with open(env_file) as irods_env:
-        irods_env_as_json =  json.load(irods_env)
-        if 'irods_ssl_ca_certificate_file' in irods_env_as_json:
-            kwargs['ssl_context'] = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH,
-                                                               cafile=irods_env_as_json['irods_ssl_ca_certificate_file'],
-                                                               capath=None, cadata=None)
+    if env_file:
+        with open(env_file) as irods_env:
+            irods_env_as_json =  json.load(irods_env)
+            if 'irods_ssl_ca_certificate_file' in irods_env_as_json:
```
trel, line 316: consider the case where someone also has `"irods_ssl_verify_server": "none"`

https://github.com/irods/python-irodsclient/issues/281

https://github.com/irods/python-irodsclient/commit/d2a5d9ff2937d42db512d706eea76efdf7037039

maybe you get it for free... with that PRC edit.

alanking: Added in new commit. Is that right?

trel: no?   `ssl_verify_server`, not the presence of the ca file only.

alanking: Oh, I misunderstood. Yet another commit.

trel: yes, i believe this will do what we want.   @d-w-moore please have a look.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -310,12 +310,15 @@ def irods_session(handler_module, meta, logger, **options):
...
-        if 'irods_ssl_ca_certificate_file' in irods_env_as_json:
-            kwargs['ssl_context'] = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH,
-                                                               cafile=irods_env_as_json['irods_ssl_ca_certificate_file'],
-                                                               capath=None, cadata=None)
+    if env_file:
+        with open(env_file) as irods_env:
+            irods_env_as_json =  json.load(irods_env)
+            verify_server = irods_env_as_json.get('irods_ssl_verify_server')
+            ca_file = irods_env_as_json.get('irods_ssl_ca_certificate_file')
+            if verify_server and verify_server != 'none' and ca_file:
```
trel, line 318: maybe move ca_file to the front of this triple check.

and reorder 316/317.

trel: scratch that - this is fine.   as you were.

alanking: Yeah, I feel that a missing or 'none'-valued `irods_ssl_verify_server` setting precludes even considering the `ca_file`, so I ordered it like that. Arguably it shouldn't even be accessed unless the first two conditions are true.

trel: agreed.  but i think the logic is good for now - it will honor the intent.

d-w-moore (commented):
It looks good to me.  Still SSL compatible for  when workers share the same filesystem, and should allow workers on a separate machine to function without an ~/.irods directory

alanking:
#'d

PR #160, opened 2021-09-10
[#159] add performance benchmark test harness

alanking:
This change adds a Docker Compose project which sets up an environment
for testing performance of the ingest tool in various conditions. The
following variables can be adjusted:
    - number of files ingested (as well as subdirectory depth)
    - traffic control delay
    - number of workers (nodes x --concurrency?)


trel:
looks very nice - please # it when ready

PR #169, opened 2022-02-24
Fix tests

alanking:
Tests run and pass in the `docker/ingest-test` project. Still trying to sort out if the version bump is appropriate at this point, but Py3.5 is no longer working with the existing dependencies


trel:
looking pretty good.

fyi: https://endoflife.date/python

d-w-moore:
Looks good to me.

Issue #40, opened 2018-08-26
ingesting files names with special characters


masilamani:
- [x] Handle register sync
- [x] Handle put
- [x] Add test for register sync
- [x] Add test for put

---
Ingesting files names with special characters throws an error "UnicodeEncodeError('utf-8', 'lock:sync_file:"

Example files:

Test_?f?[?^?x?[?X?A?b?v?f?[?g??????.doc
Test_????.zip
Test_}.ppt
?e?[?u???A?_Test.txt

Registering these files using ireg works fine.

trel:
Planning to handle this by storing a surrogate logical data_object name in the catalog:
```
logical_path: <full_path_to_collection>/irods_UnicodeEncodeError_<UUID1>
physical_path: <full_path_to_directory>/irods_UnicodeEncodeError_<UUID1>
```


And storing information about the original badfilename in the AVU for that data_object:
```
A: irods::automated_ingest::UnicodeEncodeError
V: YXNkZmFzYXNkZmFzZmRzZmFzNTR3aHdnMmdnMzM0ZzIzMmhkCg==
U: python3.base64.64encode(str)
```

This will store all the original information about the badly named file in the catalog, losslessly.

Another tool can re-inflate the information when necessary.

trel:
We need to use base64 encoding of the physical path, rather than uuid, in both the logical path and the physical path in the catalog so that updates can find the same file again the next time.  The uuid is randomized and cannot be reproduced on the next run.

PR #176, opened 2022-08-01
166.m - implement character remapping

d-w-moore:


irods_capability_automated_ingest/sync_task.py
```diff
@@ -284,6 +285,9 @@ def sync_entry(self, meta, cls, datafunc, metafunc):
             else:
                 target2 = syncer.construct_path(meta2, path)
 
+            char_map_func = getattr(event_handler.get_module(),'character_map',None)
```
trel, line 288: please add a comment about why this is the way it is...

aka... if the character map is defined, use it here, and set the new target.

irods_capability_automated_ingest/char_map_util.py
```diff
@@ -0,0 +1,23 @@
+import base64
+import hashlib
+
+#translate_string = str.translate
+from .process_map import translate_string
```
trel, line 5: what have an import from an import?    can't we just add our one/two functions to utils.py?

trel (commented):
excited about a demo!

irods_capability_automated_ingest/examples/use_character_map.py
```diff
@@ -0,0 +1,13 @@
+from irods_capability_automated_ingest.core import Core 
+from irods_capability_automated_ingest.utils import Operation
+import re
+
+class event_handler(Core):
+    re_non_alphanum = re.compile('[^a-zA-Z0-9]')      
+    @staticmethod                                     
+    def character_map():                              
+        return {event_handler.re_non_alphanum:'_'}    # map any non-ascii or non-alphanumeric
```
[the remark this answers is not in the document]
d-w-moore: the beauty of this approach is internally everything is cast to an OrderedDict.  That way you *can* pass an array of tuples to the constructor but (esp if you don't care about order) you don't have to.

irods_capability_automated_ingest/char_map_util.py
```diff
@@ -0,0 +1,23 @@
...
+        hash_suffix = ''
+        if new_el != el:
+            h = hashlib.sha1()
+            h.update(el.encode('utf8'))
+            hash_suffix = '-'+base64.b32encode(h.digest()).decode('utf8')
+        t_elem.append( new_el + hash_suffix )
+    return '/'.join(t_elem)
+
+if __name__=='__main__':
+    print(encode_path('/a/b'))
```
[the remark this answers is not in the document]
d-w-moore: yes, will remove

irods_capability_automated_ingest/char_map_util.py
d-w-moore: agreed

d-w-moore:
A keen observer might notice that we're being very inefficient by always encoding the original string using SHA1 to get the "unique-making suffix"... and they'd be right.  The suffix we encode could be potentially much shorter and store only the diff between original and transformed string.... e.g. we could use something like:
```
ascii_encode = base64.b32encode # a decent default if as many as 32 alpha-nums "map to themselves"
diff = [ x[0] for x in zip(orig_objname,transformed_objname) if x[0] != x[1] ]
suffix = ascii_encode(b"".join( [u.encode('utf8') for u in diff] ))

```

irods_capability_automated_ingest/process_map.py
```diff
@@ -0,0 +1,24 @@
+import re
+import collections
+
+_regex = re.compile('')
+string_replace = lambda s,m: s.translate({ord(k):v for k,v in m.items()})
```
d-w-moore, line 5: hm... probably want an `if v is not None` in there.  Allowing the deletion of characters won't have a good effect on uniqueness.

irods_capability_automated_ingest/examples/use_character_map.py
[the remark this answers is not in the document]
trel: Gotcha - yes OrderedDict seems sufficient.

This ingest tool is Python3.7+ only, so we can rely on known good behavior of OrderedDict...

d-w-moore:
We are considering whether to remove (or deprecate?) the feature introduced in this commit: ed8d794 . (
The original issue solved by that commit is here: https://github.com/irods/irods_capability_automated_ingest/issues/40#issue-354053528 .)

It may make sense to do that now, since we are not certain how the character remapping (the subject of the discussion within this pull request) should interact with that feature....

In that issue's problem description, it is mentioned that certain characters cause a `UnicodeEncodeError` when the directory scanner (`os.scandir`) turns up a filename or path element outside the valid UTF-8 coding space. An example would be a file created with, for example, `open(b"name"+bytes([B]),'w')` where 128 <= B < 256.
In modern OS'es and Python versions, `os.scandir` seems to use the 'surrogateescape' error handler when translating such filenames, so that a filename with binary encoding `b'name\x80'` becomes `u'name\udc80'` when decoded using UTF-8 (and thus iRODS will register it without incident). This theory is corroborated by the fact that the `UnicodeEncodeError` is now seemingly impossible to reproduce on any OS or Python version supported by the automated ingest.
It's for this reason I would suggest removal of the special handling of "illegal UTF-8" introduced in ed8d794 ....Thoughts @alanking @korydraughn  @trel @SwooshyCueb

d-w-moore:
To lay a little more solid foundation for the post above, run the script here and observe the output.  All those weird final characters (byte 128-255) are translated to their Unicode surrogate codepoints \uDC80-\uDCFF , and that is shown in the output of the script.
```python
#!/usr/bin/python3
import os
import tempfile

xdir = tempfile.mkdtemp(dir=b'.')

unicode_xdir = xdir.decode('utf8')

for b in (bytes([x]) for x in range(126,256)):
    with tempfile.NamedTemporaryFile(dir=xdir,suffix=b'_'+b):
        _file = next(os.scandir(unicode_xdir))
        print(repr( _file.name ))

```

d-w-moore:
What I'm assuming, and what I suppose is the thing needs to be thoroughly tested, is that all modern filesystems and their subobjects like to be queried as Unicode (UTF-8 encoded) entities, and that when we circumvent that by directly constructing filenames as bytestrings, the appropriate characters are given the surrogate value when they don't fit with the UTF-8 encoding.

irods_capability_automated_ingest/sync_task.py
```diff
@@ -298,9 +298,12 @@ def sync_entry(self, meta, cls, datafunc, metafunc):
 
             if 'unicode_error_filename' not in meta:
                 if char_map_func:
-                    target2 = translate_path(target2,char_map_func())
-                    meta2["b64_reason"] = "character_map"
-                    meta2['b64_path_str_X'] = b64_path_str
+                    temp = translate_path(target2,char_map_func())
+                    # arrange for AVU to be attached only when logical name changes
+                    if temp != target2:
```
trel, line 303: maybe 'translated_path' rather than temp...

d-w-moore: right

README.md
```diff
@@ -100,6 +101,45 @@ These options are used at the "start" of an ingest job.
...
+            }
+        # ...
+```
+The returned dictionary, in this case, indicates that the ingest process should replace all non-alphanumeric (as
+well as non-ASCII) characters with an underscore wherever they may occur in an otherwise normally generated logical path.
+The substition process also applies to the intermediate (ie collection name) elements in a logical path, and a suffix is
+appended to affected path elements to avoid potential collisions with other remapped object names.
+
+Each key of the returned dictionary may be a (1) character, (2) tuple of characters, (3) regular expression, or (3) a callable
+which accepts a character argument and returns a boolean indicating whether that character is one that should be remapped.
```
trel, line 123: might show an example of each of these four...

trel: and number four should not be called number three

README.md
```diff
@@ -100,6 +101,45 @@ These options are used at the "start" of an ingest job.
...
+
+In the event that the order-of-substitution is significant, the method may instead return a list of key-value tuples.
+
+### UnicodeEncodeError
+Any file whose path in the filesystem whose ingest results in a UnicodeEncodeError exception being raised (e.g. by the
```
trel, line 128: let's mention that this happens regardless of event handler code, and that it will be checked/executed prior to any character_map that may be defined.

irods_capability_automated_ingest/test/test_irods_sync.py
```diff
@@ -1492,21 +1522,45 @@ def run_scan_with_event_handler(self, eh_name, job_name = DEFAULT_JOB_NAME):
...
+            if self.DETAILED_CHECK:
+                self.assert_physical_path_and_resource(session, expected_physical_path)
+                self.assert_data_object_contents(session)
+                self.assert_data_object_size(session)
+
+#---------------------------------------------
+
+class _supplementary_tests_with_bad_filename:
+
+    @unittest.skip("SYS_RESC_DOES_NOT_EXIST - index out of range on replica - DWM 09/12/2022")
```
trel, line 1563: no DWM / dates in the skipinfo

irods_capability_automated_ingest/test/test_irods_sync.py
```diff
@@ -1605,6 +1644,49 @@ def test_put_append(self):
...
+# This is the transform built into the example files:
+# (Exclusion of '/' added here to handle path-at-a-time)
+
+regex = re.compile('[^0-9a-zA-Z/]')
+
+def character_mapping_transform(this):
+        this.remapped_bad_filename = regex.sub('_',this.BAD_FILENAME.decode('utf8'))
+        this.dest_coll_path = regex.sub('_',this.dest_coll_path)
+
+# TODO base64 transform should be the same for each of the current types of test - DWM
```
trel, line 1659: no DWMs

irods_capability_automated_ingest/test/test_irods_sync.py
```diff
@@ -1605,6 +1644,49 @@ def test_put_append(self):
...
+
+def b64_calculation(this):
+    utf8_escaped_abspath = this.bad_filepath
+    this.b64_path_str = base64.b64encode(utf8_escaped_abspath)
+
+
+class Test_irods_sync_character_mapped_path(_Test_irods_sync_with_bad_filename,
+                                            unittest.TestCase):
+    # These tests will operate on a "bad filename" wherein characters are in need of re-mapping
+    BAD_FILENAME = b'test-file~with@5!non.alphas'  # maybe add some unicode - DWM
```
trel, line 1669: no DWMs

irods_capability_automated_ingest/test/test_irods_sync.py
```diff
@@ -1605,6 +1644,49 @@ def test_put_append(self):
...
+
+
+# This is the transform built into the example files:
+# (Exclusion of '/' added here to handle path-at-a-time)
+
+regex = re.compile('[^0-9a-zA-Z/]')
+
+def character_mapping_transform(this):
+        this.remapped_bad_filename = regex.sub('_',this.BAD_FILENAME.decode('utf8'))
+        this.dest_coll_path = regex.sub('_',this.dest_coll_path)
```
trel, line 1657: indentation

irods_capability_automated_ingest/test/test_irods_sync.py
```diff
@@ -1605,6 +1644,49 @@ def test_put_append(self):
...
+    EH_SUFFIX = '_using_char_map'
+    B64_CALCULATION = b64_calculation
+    CHARACTER_MAPPING_TRANSFORM = character_mapping_transform
+    LOGICAL_PATH_CALCULATION = \
+        lambda this : join(this.dest_coll_path, this.remapped_bad_filename)
+    ALLOW_LOGICAL_NAME_SUFFIX = True
+    DETAILED_CHECK = False
+    ANNOTATION_REASON = 'character_map'
+
+#-----------------------------------------------------------
```
trel, line 1679: remove

docker/ingest-test/docker-compose.yml
```diff
@@ -23,6 +23,14 @@ services:
...
     build:
       context: provider
+      args:
+              irods_version: 4.3.0-1~bionic
+              irods_version_major_minor: 4.3
+              py_version: 3
+              # For older iRODS server:
+              #irods_version: 4.2.11-1~bionic
+              #irods_version_major_minor: 4.2
+              #py_version: ""
```
trel, line 33: mergeable?   or remove

docker/ingest-test/icommands.env
```diff
@@ -4,5 +4,7 @@ IRODS_USER_NAME=rods
 IRODS_ZONE_NAME=tempZone
 IRODS_ENVIRONMENT_FILE=/irods_environment.json
 IRODS_PASSWORD=rods
-PIP_PACKAGE=git+https://github.com/irods/irods_capability_automated_ingest@main
+#PIP_PACKAGE=git+https://github.com/irods/irods_capability_automated_ingest@main
```
trel, line 7: remove dead code

docker/ingest-test/icommands.env
```diff
@@ -4,5 +4,7 @@ IRODS_USER_NAME=rods
...
+PIP_PACKAGE=git+https://github.com/d-w-moore/irods_capability_automated_ingest@166.m
 TEST_CASE=irods_capability_automated_ingest.test.test_irods_sync
+#TEST_CASE=irods_capability_automated_ingest.test.test_irods_sync.Test_irods_sync_UnicodeEncodeError.test_register
```
trel, line 10: remove dead code

irods_capability_automated_ingest/test/test_irods_sync.py
```diff
@@ -1492,21 +1522,45 @@ def run_scan_with_event_handler(self, eh_name, job_name = DEFAULT_JOB_NAME):
...
+
+        with iRODSSession(**get_kwargs()) as session:
+            self.assert_logical_path(session, allow_suffix = self.ALLOW_LOGICAL_NAME_SUFFIX )
+            self.assert_metadata_annotation(session)
+            if self.DETAILED_CHECK:
+                self.assert_physical_path_and_resource(session, expected_physical_path)
+                self.assert_data_object_contents(session)
+                self.assert_data_object_size(session)
+
+#---------------------------------------------
```
trel, line 1559: remove comment

docker/ingest-test/docker-compose.yml
```diff
@@ -23,6 +23,14 @@ services:
   irods-catalog-provider:
     build:
       context: provider
+      args:
+              irods_version: 4.3.0-1~bionic
+              irods_version_major_minor: 4.3
+              py_version: 3
```
trel, line 29: indentation

README.md
```diff
@@ -100,6 +101,45 @@ These options are used at the "start" of an ingest job.
...
+                re.compile('[^a-zA-Z0-9]'):'_'
+            }
+        # ...
+```
+The returned dictionary, in this case, indicates that the ingest process should replace all non-alphanumeric (as
+well as non-ASCII) characters with an underscore wherever they may occur in an otherwise normally generated logical path.
+The substition process also applies to the intermediate (ie collection name) elements in a logical path, and a suffix is
+appended to affected path elements to avoid potential collisions with other remapped object names.
+
+Each key of the returned dictionary may be a (1) character, (2) tuple of characters, (3) regular expression, or (3) a callable
```
trel, line 122: please include an example for all four.

and make sure the second number three is number four.

README.md
```diff
@@ -100,6 +101,62 @@ These options are used at the "start" of an ingest job.
...
+       ('\\','#','-'): '\u2093'  # any character of the tuple is replaced by a Unicode small script x
+   ```
+   3. regular expression
+   ```
+       re.compile('[\u0100-\U0010ffff]'): '_'  # any character outside of range 0-256 becomes an underscore
+   ```
+   4. callable accepting a character argument and returning a boolean
+   ```
+       (lambda c: ord(c) in range(ord('z')+1,128)): ':'  # ASCII codes above 'z' become ':'
+   ```
```
trel, line 140: i like it - please put the comment above each example

e.g.
```
# any character of the tuple is replaced by a Unicode small script x
('\\','#','-'): '\u2093'  
```

README.md
```diff
@@ -100,6 +101,62 @@ These options are used at the "start" of an ingest job.
...
+well as non-ASCII) characters with an underscore wherever they may occur in an otherwise normally generated logical path.
+The substition process also applies to the intermediate (ie collection name) elements in a logical path, and a suffix is
+appended to affected path elements to avoid potential collisions with other remapped object names.
+
+Each key of the returned dictionary indicates a character or set of characters needing substitution.
+Possible key types include:
+
+   1. character
+   ```
+       '\\':'_'	 # substitute backslashes with underscores
```
trel, line 127: space after colon?

docker/ingest-test/docker-compose.yml
d-w-moore: Mergeable and maybe important to show that tests may be run on 4.2 as well as 4.3 . Does this require more comments to make things more explicit?

korydraughn: This can be documented in a README that lives inside of docker/ingest-test.

IMO, yes this needs more words. Right now, people can only assume what these do. We need to do our best to take the guess work out of things. That doesn't mean we have to document every single thing, but providing words as to what each parameter is for and then providing an example of usage is probably good enough.

trel:
one indentation comment left

and then resolve the ones that are done.   looking pretty good.

irods_capability_automated_ingest/test/test_irods_sync.py
d-w-moore: done.

d-w-moore:
Putting it through one more test before merge.

trel:
Click resolve above on two comments and squash when ready.   No # yet.

d-w-moore:
Squashed with no #

irods_capability_automated_ingest/sync_task.py
```diff
@@ -284,7 +291,23 @@ def sync_entry(self, meta, cls, datafunc, metafunc):
             else:
                 target2 = syncer.construct_path(meta2, path)
 
+            # If the event handler has a character_map function, it should have returned a
+            # structure (either a dict or a list/tuple of key-value tuples) to be used for
```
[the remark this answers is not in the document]
d-w-moore: It won't be a silent failure , but it won't be pretty either. Should  an explicit error could be thrown, which might make things friendlier?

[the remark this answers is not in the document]
trel: Yes, we should check for a sane/valid value and STOP any work being done if it's not.

The error should say that the return type of the character_map was invalid.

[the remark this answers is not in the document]
d-w-moore: Sounds like we should check that up front then, at the moment the --event_handler option is submitted to the `irods_sync.py` script. Sound like the right approach, @alanking ?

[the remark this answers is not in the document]
korydraughn: If we can detect issues on startup, then the rest of the code becomes faster because it can assume things are in a good state. That also means the rest of the code must be implemented so that invalid values do not appear as a result of carrying out various operations.

So overall, yes.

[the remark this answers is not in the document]
d-w-moore: I've made an attempt, want to look it over?  Have tested manually, too.  Does it need its own unit test ?

irods_capability_automated_ingest/irods_sync.py
```diff
@@ -76,6 +101,8 @@ def handle_start(args):
     if ex_file_arg != None:
         ex_arg_list = [x.strip() for x in ex_file_arg[0].split(',')]
 
+    check_event_handler(args.event_handler)
```
trel, line 104: this will run once per sync job, not once per file-to-be-synced, yes?

irods_capability_automated_ingest/irods_sync.py
```diff
@@ -43,6 +43,31 @@ def get_celery_broker_info():
     return host, port, db
 
 
+class character_map_argument_error(Exception): pass
+
+def check_event_handler(filename):
```
trel, line 48: please put the comment from the commit message here in the code...
- the goal, the reason, the what.

irods_capability_automated_ingest/sync_task.py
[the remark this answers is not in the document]
trel: left comments

irods_capability_automated_ingest/irods_sync.py
d-w-moore: Right. Actually at the command line invocation level, before anything starts, an exception will be thrown if the named event handler is not  correct.

irods_capability_automated_ingest/sync_task.py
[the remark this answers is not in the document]
d-w-moore: Good.  And yes, tests passing as designed on 4.2.11 and 4.3.0

irods_capability_automated_ingest/irods_sync.py
trel: perfect.

trel:
and squash again when ready.  still no #.

d-w-moore:
Ok, done.

irods_capability_automated_ingest/irods_sync.py
d-w-moore: yes

.gitignore
```diff
@@ -0,0 +1,3 @@
+*.pyc
+*egg-info/
+cscope.*
```
trel, line 3: maybe create a housekeeping issue for docs/typos/configs

d-w-moore: ok, done.

trel (approved):
looks good to me.  let's await @alanking's eyes one last time before #.

alanking (approved):

trel:
alright dan, please # em.

d-w-moore:
done!

Issue #129, opened 2020-03-17
PUT_SYNC does not work with S3


acnewton:
**What we want**
Sync data from AWS S3 bucket to iRODS using the PUT_SYNC operation.

**What we did**
`python -m irods_capability_automated_ingest.irods_sync start --ignore_cache --event_handler 'irods_capability_automated_ingest.examples.sync' --synchronous --progress --s3_keypair aws-s3-keypair --s3_region_name eu-west-1 --log_filename /home/irods/log/test.log --log_level DEBUG /bucket-name /tempZone/home/rods/target
 [Elapsed Time: 0:00:01] |####################################################################################################################################################| (Time:  0:00:01) count:      1 tasks: ------ failures:      1 retries: ------
(rodssync) 
13:29 myhost:/home/irods
`

**What we expected**
Successful synchronization: file copied from S3 bucket to iRODS target collection.

**What we got**
 Error log that the file can not be found locally.

**Possible solution**
It seems that functionality to download/stream data first from S3 in upload_file() and sync_file() functions in irods_capability_automated_ingest/sync_irods.py is missing.

zfed:
we experimented same issue with operation.PUT 
there are no solutions yet?

alanking:
This work was done a while ago, but I dropped the ball on getting it reviewed: https://github.com/irods/irods_capability_automated_ingest/pull/134 The work still needs to be rebased (a lot has moved since then) and tested.

avkrishnamurthy:
We have brainstormed several potential solutions to solve this issue. The current plan is to try to implement #**2**, and subsequently #**1**, which will be the most efficient option. Currently, #**4**, is already hypothetically working thanks to acnewton's pull request from 2020, which will be the fallback option if #**1**, or #**2** don't work. 
 ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ----
**1. Multi-read from S3 -> Multi-write to iRODS**
- Pros:
   - Best parallelization
   - Reduced bottlenecks with multiple buffers
   - Scalability
- Cons:
   - More complex to implement
   - May use more resources for multiple buffers and multiple threads
 ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ----
**2. Single stream/read from S3 -> Multi-write to iRODS**
- Pros:
   - Probably easier to do multi-write when putting to iRODS than getting from S3
   - Stepping stone to multi-read, multi-write
[...]

PR #206, opened 2023-06-30
[#129] single stream put from S3

avkrishnamurthy:
This fixes the "File not found error" that occurs when trying to run a PUT or PUT_SYNC for files from an S3 resource. I am using Minio to get the file from the S3 bucket. Currently can not use parallelization to do the put for S3 objects, which will need to be worked on. Not sure what to do with that at this point.  I have also included checksum comparison with the Etag for both normal files in an S3 bucket and also multipart upload files. I am not sure if this is the correct way to do the checksum, so feedback there would be good. The default chunk size for multipart uploads to Amazon S3 is 8 MB, but I have read online that it may not always be the same, so I have added a command line argument to the ingest job to specify a different chunk size.

### DIFF
--- a/irods_capability_automated_ingest/irods_sync.py
+++ b/irods_capability_automated_ingest/irods_sync.py
@@ -130,6 +130,7 @@ def handle_start(args):
     data["s3_keypair"] = args.s3_keypair
     data["s3_proxy_url"] = args.s3_proxy_url
     data["s3_secure_connection"] = not args.s3_insecure_connection
+    data["s3_multipart_chunksize_in_mib"] = args.s3_multipart_chunksize_in_mib
     data["exclude_file_type"] = ex_arg_list
     data['exclude_file_name'] = [''.join(r) for r in args.exclude_file_name]
     data['exclude_directory_name'] = [
@@ -198,6 +199,9 @@ def main():
                               type=str, default=None, help='URL to proxy for S3 access')
     parser_start.add_argument('--s3_insecure_connection', action="store_true",
                               default=False, help='Do not use SSL when connecting to S3 endpoint')
+    parser_start.add_argument('--s3_multipart_chunksize_in_mib',action="store", type=int,
+                              default=8, choices=range(5, 5001), metavar="[5-5000]",
+                              help='Chunk size in mebibytes for multipart S3 uploads. Minimum part size is 5 MiB and the maximum part size is 5000 MiB.')
     parser_start.add_argument('--exclude_file_type', nargs=1, action="store", default='none',
                               help='types of files to exclude: regular, directory, character, block, socket, pipe, link')
     parser_start.add_argument('--exclude_file_name', type=list, nargs='+', action="store", default='none',
--- a/irods_capability_automated_ingest/scanner.py
+++ b/irods_capability_automated_ingest/scanner.py
@@ -165,9 +165,8 @@ class s3_scanner(scanner):
         endpoint_domain = meta.get('s3_endpoint_domain')
         s3_access_key = meta.get('s3_access_key')
         s3_secret_key = meta.get('s3_secret_key')
-        s3_secure_connection = meta.get('s3_secure_connection')
-        if s3_secure_connection is None:
-            s3_secure_connection = True
+        s3_secure_connection = meta.get('s3_secure_connection', True)
+       
         client = Minio(
             endpoint_domain,
             access_key=s3_access_key,
--- a/irods_capability_automated_ingest/sync_actions.py
+++ b/irods_capability_automated_ingest/sync_actions.py
@@ -106,19 +106,14 @@ def start_job(data):
     s3_region_name = data["s3_region_name"]
     s3_endpoint_domain = data["s3_endpoint_domain"]
     s3_keypair = data["s3_keypair"]
-
+    s3_multipart_chunksize = data["s3_multipart_chunksize_in_mib"]
     logger = sync_logging.get_sync_logger(logging_config)
     data_copy = data.copy()
 
     if s3_keypair is not None:
-        data_copy['s3_region_name'] = s3_region_name
-        data_copy['s3_endpoint_domain'] = s3_endpoint_domain
-        data_copy['s3_keypair'] = s3_keypair
-        # parse s3 keypair
-        if s3_keypair is not None:
-            with open(s3_keypair) as f:
-                data_copy['s3_access_key'] = f.readline().rstrip()
-                data_copy['s3_secret_key'] = f.readline().rstrip()
+        with open(s3_keypair) as f:
+            data_copy['s3_access_key'] = f.readline().rstrip()
+            data_copy['s3_secret_key'] = f.readline().rstrip()
         # set source
         src_abs = src_path
     else:
--- a/irods_capability_automated_ingest/sync_irods.py
+++ b/irods_capability_automated_ingest/sync_irods.py
@@ -6,13 +6,15 @@ from irods.exception import NetworkException
 from .sync_utils import size, get_redis
 from .utils import Operation
 from .custom_event_handler import custom_event_handler
-import redis_lock
 from minio import Minio
+import redis_lock
 import json
 import irods.keywords as kw
 import base64
+import hashlib
 import ssl
 import threading
+
 
 def validate_target_collection(meta, logger):
     # root cannot be the target collection
@@ -95,6 +97,7 @@ def annotate_metadata_for_special_data_objs(meta, session, source_physical_fullp
             os.path.join(os.path.dirname(source_physical_fullpath), os.readlink(source_physical_fullpath)),
             'automated_ingest')
 
+
 def register_file(hdlr_mod, logger, session, meta, **options):
     dest_dataobj_logical_fullpath = meta["target"]
     source_physical_fullpath = meta["path"]
@@ -121,7 +124,7 @@ def register_file(hdlr_mod, logger, session, meta, **options):
     options[kw.DATA_SIZE_KW] = str(meta['size'])
     options[kw.DATA_MODIFY_KW] = str(int(meta['mtime']))
 
-    logger.info("registering object " + dest_dataobj_logical_fullpath + ", options = " + str(options))
+    logger.info("registering object " + dest_dataobj_logical_fullpath + ", source_physical_fullpath: " + source_physical_fullpath+ ", options = " + str(options))
     session.data_objects.register(phypath_to_register_in_catalog, dest_dataobj_logical_fullpath, **options)
 
     logger.info("succeeded", task="irods_register_file", path = source_physical_fullpath)
@@ -149,9 +152,8 @@ def get_s3_client(meta):
     s3_region_name = meta.get('s3_region_name')
     s3_access_key = meta.get('s3_access_key')
     s3_secret_key = meta.get('s3_secret_key')
-    s3_secure_connection = meta.get('s3_secure_connection')
-    if s3_secure_connection is None:
-        s3_secure_connection = True
+    s3_secure_connection = meta.get('s3_secure_connection', True)
+
     client = Minio(
                  endpoint_domain,
                  region=s3_region_name,
@@ -161,18 +163,20 @@ def get_s3_client(meta):
                  http_client=httpClient)
     return client
 
-
 def upload_file_from_S3(logger, session, meta, src, dest, offset=0, **options):
 
-    BUFFER_SIZE = 1024
+    #Could pass in as parameter when running ingest job, keep in meta
+    #May be needed for files uploaded to AWS S3 with multi-part upload
+    BUFFER_SIZE = 1024*1024
     client = get_s3_client(meta)
     path_list = src.lstrip('/').split('/', 1)
     if len(path_list) == 1:
-        logger.error("failed", task="irods_upload_S3file", path=src)
+        raise ValueError("Object name or bucket name missing from path to S3 object: [{}]".format(src))
     else:
         bucket_name = path_list[0]
         object_name = path_list[1]
 
+    #Pulled from put function from data_object_manager in PRC
     #Set operation type to trigger acPostProcForPut
     if kw.OPR_TYPE_KW not in options:
         options[kw.OPR_TYPE_KW] = 1 #PUT_OPR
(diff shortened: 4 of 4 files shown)


korydraughn:
In the future, please leave off the pounds. We only add them once the changes have been reviewed and approved.

The reason for this is that if more changes are needed due to the review, pushing more commits causes the issue to be flooded with info about those new commits. So leaving the pounds off guards against that.

trel:
if this PR is based on top of #134 then we should just add new commits, to preserve credit.

d-w-moore:
Other than reaffirming @trel's statement above, this otherwise seems fine.  I gather that it is a temporary substitute for the "complete" PUT from S3 solution we discussed yesterday, which will ultimately be multi-read/multi-write (NxN threads) from S3 to iRODS ?

avkrishnamurthy:
> Other than reaffirming @trel's statement above, this otherwise seems fine. I gather that it is a temporary substitute for the "complete" PUT from S3 solution we discussed yesterday, which will ultimately be multi-read/multi-write (NxN threads) from S3 to iRODS ?

Yes, this is just the initial naive solution using just single stream to stream.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -12,6 +12,8 @@
 import base64
 import ssl
 import threading
+import hashlib
+import minio
```
d-w-moore, line 16: Could alphabetize.

d-w-moore (approved):

trel:
and i'm happy to get this merged as stream-to-stream, and then we can optimize in a new issue.

alanking:
We are working on getting a branch which preserves the original authorship and builds on top of it. Stand by...

trel:
oooh, git school is in.

alanking:
Force-push seems to have been a success.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -120,18 +124,129 @@ def register_file(hdlr_mod, logger, session, meta, **options):
...
+    s3_access_key = meta.get('s3_access_key')
+    s3_secret_key = meta.get('s3_secret_key')
+    s3_secure_connection = meta.get('s3_secure_connection')
+    if s3_secure_connection is None:
+        s3_secure_connection = True
+
+    isLocalHost = ['localhost', '127.0.0.1']
+    #Minio server running locally causes SSL error if secure set to True
+    if endpoint_domain.split(":")[0] in isLocalHost:
+        s3_secure_connection = False
```
alanking, line 162: Same comment as before here - I think this should be specified by the framework, not automatic.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -120,18 +124,129 @@ def register_file(hdlr_mod, logger, session, meta, **options):
...
+                                    status_forcelist=[500, 502, 503, 504]
+                                )
+                     )
+    endpoint_domain = meta.get('s3_endpoint_domain')
+    s3_region_name = meta.get('s3_region_name')
+    s3_access_key = meta.get('s3_access_key')
+    s3_secret_key = meta.get('s3_secret_key')
+    s3_secure_connection = meta.get('s3_secure_connection')
+    if s3_secure_connection is None:
+        s3_secure_connection = True
```
alanking, line 157: This could be:
```suggestion
    s3_secure_connection = meta.get('s3_secure_connection', True)
```

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -120,18 +124,129 @@ def register_file(hdlr_mod, logger, session, meta, **options):
...
+
+def upload_file_from_S3(logger, session, meta, src, dest, offset=0, **options):
+
+    #Could pass in as parameter when running ingest job, keep in meta
+    #May be needed for files uploaded to AWS S3 with multi-part upload
+    BUFFER_SIZE = 1024
+    client = get_s3_client(meta)
+    path_list = src.lstrip('/').split('/', 1)
+    if len(path_list) == 1:
+        logger.error("failed", task="irods_upload_S3file", path=src)
```
alanking, line 181: Do we need to bail here? What does the error mean if it's allowed to continue?

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -120,18 +124,129 @@ def register_file(hdlr_mod, logger, session, meta, **options):
...
+    etag = response.getheader('Etag').replace('"','')
+    recreated_etag = None
+    #Doing the put here because session.data_objects.put will not recognize the file/directory (aka bucket) since it is not contained on the local fs
+    #Why we use minio to get the data object from the bucket and to write the contents to the destination data object
+    #Also calculating checksum 
+    #Preliminary way of determining if file is multipart or not; etag from amazonS3 will be >32 for multipart, 32 for regular
+    if len(etag)>32:
+        logger.info("Multipart object", task="irods_S3upload_file")
+        c_size = meta.get('s3_multipart_chunksize')
+        if not c_size: c_size=8
```
alanking, line 212:
```suggestion
        c_size = meta.get('s3_multipart_chunksize', 8)
```

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -120,18 +124,129 @@ def register_file(hdlr_mod, logger, session, meta, **options):
...
+        hash_md5 = hashlib.md5()
+        for data in response.stream(amt=BUFFER_SIZE):
+            hash_md5.update(data)
+            tfd.write(data)
+        tfd.close()
+        recreated_etag = hash_md5.hexdigest()
+
+    response.release_conn()
+    if etag!=recreated_etag:
+        logger.info("checksums do not match! etag {} recreated etag {}".format(etag,recreated_etag), task="irods_S3upload_file", path=src)
```
alanking, line 244: I feel like this is more like a `warn` or an `error`... Thoughts?

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -168,28 +287,42 @@ def sync_file(hdlr_mod, logger, session, meta, **options):
...
-                if buf == b"":
-                    break
-                tfd.write(buf)
-        tfd.close()
-        logger.info("succeeded", task="irods_append_file", path=source_physical_fullpath)
-
+    s3_keypair = meta.get("s3_keypair")
+    logger.info("checking if data object from S3 " + str(s3_keypair))
+    if op is None:
+        op = Operation.REGISTER_SYNC
```
alanking, line 293: `sync_file` is for `PUT_SYNC` or `PUT_APPEND` operations. `op` being `None` should either be treated as an error or we should default to `PUT_SYNC`. Thoughts?

irods_capability_automated_ingest/sync_irods.py
avkrishnamurthy: Yes, I would think we need to bail. If the length of the path list is less than 1, it means that the source_physical_fullpath either does not contain the filename within the S3 bucket, or does not contain the bucket name. If we continue, it would lead to an error when trying to get the object from the S3 bucket via the Minio client. 

What does it look like to bail from the upload file? I'm not sure what needs to happen.

irods_capability_automated_ingest/scanner.py
```diff
@@ -168,6 +168,11 @@ def instantiate(self, meta):
         s3_secure_connection = meta.get('s3_secure_connection')
         if s3_secure_connection is None:
             s3_secure_connection = True
+       
+        isLocalHost = ['localhost', '127.0.0.1']
+        #Minio server running locally causes SSL error if secure set to True
+        if endpoint_domain.split(":")[0] in isLocalHost:
+            s3_secure_connection = False
```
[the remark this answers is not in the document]
avkrishnamurthy: Yes, that is definitely the better option. I totally forgot that it was there.

irods_capability_automated_ingest/sync_irods.py
avkrishnamurthy: I'm not totally sure. I don't fully know what purpose making the op REGISTER_SYNC serves, I think I mostly included it because it is there in sync_file in main (I attached a screenshot). I think it makes more sense to have it default to PUT_SYNC?
image

irods_capability_automated_ingest/sync_actions.py
```diff
@@ -106,14 +106,15 @@ def start_job(data):
...
-
+    s3_multipart_chunksize = data["s3_multipart_chunksize"]
     logger = sync_logging.get_sync_logger(logging_config)
     data_copy = data.copy()
 
     if s3_keypair is not None:
         data_copy['s3_region_name'] = s3_region_name
         data_copy['s3_endpoint_domain'] = s3_endpoint_domain
         data_copy['s3_keypair'] = s3_keypair
+        data_copy['s3_multipart_chunksize']=s3_multipart_chunksize
```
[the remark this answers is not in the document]
avkrishnamurthy: Yeah, they should probably be removed. I was just looking for the places where the other s3 options were so that I could cover all my bases when adding the multipart chunksize option. I'll try removing it and running the ingest again to make sure it doesn't break anything.

irods_capability_automated_ingest/irods_sync.py
```diff
@@ -198,6 +199,8 @@ def main():
                               type=str, default=None, help='URL to proxy for S3 access')
     parser_start.add_argument('--s3_insecure_connection', action="store_true",
                               default=False, help='Do not use SSL when connecting to S3 endpoint')
+    parser_start.add_argument('--s3_multipart_chunksize',action="store", type=int,
```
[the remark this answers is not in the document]
avkrishnamurthy: Yes, I like that idea. 

https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html

This says the min - max part sizes are 5 MiB to 5 GiB. Pretty wide range, but maybe we can just give the range as [5-5000] MB? (Also not sure about what to do with MiB vs MB)

irods_capability_automated_ingest/sync_irods.py
avkrishnamurthy: Yes, I'm thinking it should be an error since it means that the file was not transferred successfully.

irods_capability_automated_ingest/irods_sync.py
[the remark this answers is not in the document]
alanking: Ah, okay. If we want to be more precise, then we can say `_in_mib` or `_in_mebibytes`. I will defer.

irods_capability_automated_ingest/sync_irods.py
alanking: I think we need to raise an Exception. I think a `ValueError` would be appropriate here based on your description.

On that note, we should definitely leave a more detailed message here than "failed" as well. Perhaps something to the effect of your explanation would suffice.

irods_capability_automated_ingest/sync_irods.py
alanking: Right. I think `REGISTER_SYNC` is just incorrect here. So, if there's no operation it should either be an error or default to something else. I'm leaning toward raising an Exception here as the operation should not be empty by this point. Thoughts?

irods_capability_automated_ingest/irods_sync.py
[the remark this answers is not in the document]
alanking: Oops, didn't answer the question. Yes, that range seems fine to me. It's a big range, but presents no chance of confusion.

irods_capability_automated_ingest/sync_irods.py
avkrishnamurthy: Yes, makes sense

alanking: I think that it would be okay to just make a new issue and look into this later. After all, this is not *new*, as you pointed out, and the implications of the change are unknown at this point. Thoughts?

korydraughn: I agree. Let's capture this in an issue and come back to it later.

avkrishnamurthy: Before I saw these replies, in my last commit I added code to raise an Exception here if the operation is None. Should I change this back to setting it to be ```REGISTER_SYNC```?

irods_capability_automated_ingest/irods_sync.py
```diff
@@ -199,8 +199,9 @@ def main():
                               type=str, default=None, help='URL to proxy for S3 access')
     parser_start.add_argument('--s3_insecure_connection', action="store_true",
                               default=False, help='Do not use SSL when connecting to S3 endpoint')
-    parser_start.add_argument('--s3_multipart_chunksize',action="store", type=int,
-                              default=8, help='File chunk size for multi-part upload to Amazon S3 (MB)')
+    parser_start.add_argument('--s3_multipart_chunksize_in_mb',action="store", type=int,
+                              default=8, choices=range(5, 5001), metavar="[5-5000]",
+                              help='File chunk size in megabytes for files uploaded multi-part to Amazon S3 (MB). Minimum part size specified by AWS is 5 MB, and the maximum part size is 5000 MB.')
```
alanking, line 204: I think we should avoid using "Amazon" and "AWS" here as multi-part is not exclusive to Amazon's S3 but is part of the S3 protocol. Thoughts?

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -174,11 +167,11 @@ def upload_file_from_S3(logger, session, meta, src, dest, offset=0, **options):
 
     #Could pass in as parameter when running ingest job, keep in meta
     #May be needed for files uploaded to AWS S3 with multi-part upload
-    BUFFER_SIZE = 1024
+    BUFFER_SIZE = 1024*1024
     client = get_s3_client(meta)
     path_list = src.lstrip('/').split('/', 1)
     if len(path_list) == 1:
-        logger.error("failed", task="irods_upload_S3file", path=src)
+        raise ValueError("Incorrect path to S3 object. Object name or bucket name missing.")
```
alanking, line 174: Maybe include src in this as well so the user can tell which path was busted?
```suggestion
        raise ValueError("Object name or bucket name missing from path to S3 object: [{}]".format(src))
```

irods_capability_automated_ingest/sync_irods.py
alanking: Yes, I think putting it back would be safest for now. We could also leave a TODO comment linked to the new issue so that it's flagged.

alanking (commented):
Getting closer...

avkrishnamurthy:
> Getting closer...

Added changes

irods_capability_automated_ingest/irods_sync.py
```diff
@@ -199,9 +199,9 @@ def main():
                               type=str, default=None, help='URL to proxy for S3 access')
     parser_start.add_argument('--s3_insecure_connection', action="store_true",
                               default=False, help='Do not use SSL when connecting to S3 endpoint')
-    parser_start.add_argument('--s3_multipart_chunksize_in_mb',action="store", type=int,
+    parser_start.add_argument('--s3_multipart_chunksize_in_mib',action="store", type=int,
                               default=8, choices=range(5, 5001), metavar="[5-5000]",
-                              help='File chunk size in megabytes for files uploaded multi-part to Amazon S3 (MB). Minimum part size specified by AWS is 5 MB, and the maximum part size is 5000 MB.')
+                              help='File chunk size in mebibytes for files uploaded multi-part to S3. Minimum part size is 5 MiB, and the maximum part size is 5000 MiB.')
```
alanking, line 204: What do you think of this rewording?
```suggestion
                              help='Chunk size in mebibytes for multipart S3 uploads. Minimum part size is 5 MiB and the maximum part size is 5000 MiB.')
```
I think the "uploaded multi-part" wording is tripping me up, so any kind of rewording there would seem better to me.

alanking (commented):
Some parting thoughts from me.

@d-w-moore / @JustinKyleJames - Any concerns on the python or S3 side of things?

irods_capability_automated_ingest/irods_sync.py
avkrishnamurthy: Yeah, let's go with that.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -278,8 +278,11 @@ def sync_file(hdlr_mod, logger, session, meta, **options):
 
     logger.info("syncing object " + dest_dataobj_logical_fullpath + ", options = " + str(options))
     op = event_handler.operation(session, **options)
+
+    # TODO: If op is None, we might want to raise an exception as this function should only be called for PUT_SYNC and PUT_APPEND...
```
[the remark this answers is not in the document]
avkrishnamurthy: image

Something like this?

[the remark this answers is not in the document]
avkrishnamurthy: image

Something like this?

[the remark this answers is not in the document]
alanking: Yep, perfect.

alanking (commented):
Everything looks good to me!

Perhaps we squash this down to 2-3 commits (omitting #s for now) and have one last look.

JustinKyleJames:
This looks fine to me as long as it has been tested and performs as expected.

avkrishnamurthy:
> Everything looks good to me!
> 
> Perhaps we squash this down to 2-3 commits (omitting #s for now) and have one last look.

squashed

alanking (approved):
Looks good to me. Let's add the #s

avkrishnamurthy:
> Looks good to me. Let's add the #s

I added the pounds to the 2 commits and to the PR title. Let me know if I need to change anything else.

PR #209, opened 2023-07-19
[ 207] multi read and write from S3 to iRODS for put, putsync

avkrishnamurthy:
The ```PUT``` and ```PUT_SYNC``` bug for S3 objects were resolved in #206, but still run very slowly because the solution uses a single stream read and write from an S3 object to an iRODS collection. For larger files, this is a problem. This PR speeds up this process using multithreading to do parallel read and write from S3 into iRODS.


irods_capability_automated_ingest/sync_irods.py
```diff
@@ -131,50 +133,99 @@ def register_file(hdlr_mod, logger, session, meta, **options):
...
+    num_threads = max(1, min(multiprocessing.cpu_count(), RECOMMENDED_NUM_THREADS_PER_TRANSFER))
+    open_options = {kw.NUM_THREADS_KW: str(num_threads), kw.DATA_SIZE_KW:str(total_bytes_)}
+    if target_resc_name:
+        open_options[kw.RESC_NAME_KW] = target_resc_name
+        open_options[kw.DEST_RESC_NAME_KW] = target_resc_name
+    (Io, rawfile) = session.data_objects.open_with_FileRaw(dest_, 'w', finalize_on_close=True, **open_options)
+    (replica_token, resc_hier) = rawfile.replica_access_info()
+    bytes_per_thread = total_bytes_//num_threads
+    chunk_ranges = [range(j * bytes_per_thread, (j+1)*bytes_per_thread) for j in range(num_threads-1)]
+    chunk_ranges.append(range((num_threads-1)*bytes_per_thread, total_bytes_))
```
alanking, line 203: These lines could benefit from an explanatory comment.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -131,50 +133,99 @@ def register_file(hdlr_mod, logger, session, meta, **options):
...
+    num_threads = min(num_threads, len(chunk_ranges))
+    mgr = _Multipart_close_manager(Io, threading.Barrier(num_threads))
+    counter = 1
+    ops = {}
+    ops[kw.NUM_THREADS_KW]=str(num_threads)
+    ops[kw.DATA_SIZE_KW]=str(total_bytes_)
+    ops[kw.RESC_HIER_STR_KW]=resc_hier
+    ops[kw.REPLICA_TOKEN_KW]=replica_token
+
+    for byte_range in chunk_ranges:
```
alanking, line 215: Optionally, you could use `enumerate(chunk_ranges)` and get a `counter` for free.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -286,13 +358,15 @@ def sync_file(hdlr_mod, logger, session, meta, **options):
...
-    
+
 
     if s3_keypair:
         if op == Operation.PUT_APPEND:
             logger.info("appending object " + source_physical_fullpath + "from S3, options = " + str(options))
             tsize = size(session, dest_dataobj_logical_fullpath)
-            upload_file_from_S3(logger, session, meta, source_physical_fullpath, dest_dataobj_logical_fullpath, offset=tsize, **options)
+            client = MinioClient.get_instance(meta)
+            bucket_name, object_name = get_bucket_object_names_S3(source_physical_fullpath)
```
alanking, line 368: Why split these out here when we can just pass the source path to `single_stream_upload_from_S3` and have it do it for us? Not a rhetorical question. :)

alanking (commented):
A fine start.

Consider breaking out some of this stuff into a new file and separating the concerns a bit. We created the concept of a generic "scanner" with implementations for a filesystem and an S3 bucket. Perhaps the S3 scanner can call these new "upload" functions and we won't need as much if/else in here.

Adding @d-w-moore as a reviewer for expert review of the use of the parallel transfer mechanisms of PRC.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -131,50 +133,99 @@ def register_file(hdlr_mod, logger, session, meta, **options):
...
-    return client
+class MinioClient:
+    _instance = None
 
-def upload_file_from_S3(logger, session, meta, src, dest, offset=0, **options):
+    @classmethod
+    def get_instance(cls, meta):
+        if cls._instance is None:
+            cls._instance = cls._create_instance(meta)
+        return cls._instance
```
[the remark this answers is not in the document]
avkrishnamurthy: I thought to make a Singleton because originally a new client object was being created every time the upload_file_from_S3 function is called, but this client object will never change for the entirety of the ingest job, so I thought it doesn't make sense to create a new object every time. Should I change it back? I am reading online now that Singletons seem to be very widely disliked.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -131,50 +133,99 @@ def register_file(hdlr_mod, logger, session, meta, **options):
...
+                     endpoint_domain,
+                     region=s3_region_name,
+                     access_key=s3_access_key,
+                     secret_key=s3_secret_key,
+                     secure=s3_secure_connection,
+                     http_client=httpClient)
+        return client
+
+COPY_BUF_SIZE = (1024**2) * 4
+def thread_copy(objHandle, range_, client_, bucket_name_, object_name_, mgr_, thread_debug_id):
```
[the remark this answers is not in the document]
avkrishnamurthy: No real reason, I saw it was being done in PRC, so I guess I was kinda following that. It probably makes more sense to not include it.

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -225,15 +273,39 @@ def upload_file_from_S3(logger, session, meta, src, dest, offset=0, **options):
...
+    if etag!=recreated_etag and offset==0:
+        logger.error("checksums do not match! etag {} recreated etag {}".format(etag,recreated_etag), task="irods_S3upload_file", path="/"+bucket_name_+"/"+object_name_)
+
+def get_bucket_object_names_S3(src):
+    path_list = src.lstrip('/').split('/', 1)
+    if len(path_list)==1:
+        raise ValueError("Object name or bucket name missing from path to S3 object: [{}]".format(src))
+    return path_list[0], path_list[1]
+
+def upload_file_from_S3(logger, session, meta, src, dest, offset=0, **options):
```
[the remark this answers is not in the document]
avkrishnamurthy: It is essentially just the starting point for the S3 put. I split it up into parallel_upload_from_S3 and single_stream_upload_from_S3, but this function determines if the file should be uploaded parallel or single stream, calls the respective function and checks that the upload went smoothly. I wanted to keep it because it makes upload_file a little cleaner by not having to get the minio client, check the file size and target resource by offloading it to this function.

irods_capability_automated_ingest/sync_irods.py
avkrishnamurthy: No good reason. Mostly did it like that from laziness having to change parameters and add the ```get_bucket_object_names_S3``` in ```parallel_upload_from_S3``` and ```single_stream_upload_from_S3```. I agree that it will be more sensible to just call ```single_stream_upload_from_S3``` with the source path, and same for parallel upload.

avkrishnamurthy:
> A fine start.
> 
> Consider breaking out some of this stuff into a new file and separating the concerns a bit. We created the concept of a generic "scanner" with implementations for a filesystem and an S3 bucket. Perhaps the S3 scanner can call these new "upload" functions and we won't need as much if/else in here.
> 
> Adding @d-w-moore as a reviewer for expert review of the use of the parallel transfer mechanisms of PRC.

You are referring to the scanner.py, correct?

irods_capability_automated_ingest/sync_irods.py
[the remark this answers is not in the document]
alanking: Oh, I see. That seems like a valid reason to do this, especially if there are performance implications. I think it's okay to keep it unless anybody is vehemently opposed.

irods_capability_automated_ingest/sync_irods.py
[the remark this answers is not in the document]
alanking: Oh, so is the call on L369 to `single_stream_upload_from_S3` supposed to be `upload_file_from_S3`?

irods_capability_automated_ingest/sync_irods.py
alanking: Okay. If you think it makes sense to split up the bucket name and object name for the parameters, I'm willing to hear it out. Just wondering!

alanking:
Yes, scanner.py. We may be able to have the scanner call these functions directly instead of routing things through `upload_file` and (re-)determining if it's an S3 path or a filesystem path. I'm open to whatever way makes sense, though.

irods_capability_automated_ingest/sync_irods.py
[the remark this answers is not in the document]
avkrishnamurthy: Yeah, now that I'm looking at it, it makes much more sense that it should be calling ```upload_file_from_S3```. I even added the check that ```offset==0``` in ```upload_file_from_S3``` to make sure it doesn't try to do the ```PUT_APPEND``` in parallel, but I must have just forgotten to change it to ```upload_file_from_S3``` in ```sync_file```.

irods_capability_automated_ingest/sync_irods.py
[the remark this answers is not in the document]
korydraughn: I agree.

The singleton pattern just requires more care to avoid bad situations.

Is the singleton shared across multiple jobs?
How many things need access to it?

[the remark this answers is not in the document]
avkrishnamurthy: If I am understanding how the celery workers operate, each celery worker has a set of tasks that it runs (like getting the paths of the files in the to-ingest directory/bucket, checking mtimes) that will eventually call the ```upload_file``` function for each file in its batch of files (or something else like ```register_file```, depending on the operation type from the event_handler). 

For S3 files, this will then call the ```upload_file_from_S3``` function, which will get the instance of the Minio client object. This means that multiple celery workers could potentially be accessing the instance of the Minio client object at the same time when they each call the ```upload_file_from_S3``` function for a file in their batch of files, which might be an issue? But since this instance of the client object is just being used to access files in the S3 bucket, I don't think it should pose any problems, theoretically. 

Next in ```upload_file_from_S3```, if the file is large enough, the upload will be parallelized, and I pass a reference to this instance of the Minio client to each of the n-threads, which each thread uses to get a chunk of the S3 object. There should be no concurrency issues with this because each thread is accessing a different part of the S3 object and Minio is threadsafe according to their docs. 

The reason that I wanted to use a Singleton was because I thought it could help with performance, and it made more sense to me logically that instead of creating a new instance of the Minio client object every time an individual celery worker's task calls the upload_file_from_S3 function, the same instance is reused because the Minio client will always be the same (just requires the endpoint domain, and secret key and password). Originally, my plan was to figure out a way to utilize the Scanner object created in the ```sync_path``` task in ```sync_task.py```, which already creates a Minio client to get paths to the objects in the to-ingest S3 bucket. I wanted to pass this object beyond the scope of the ```sync_path``` task, but it seems like it will require some changes to how the scanner object in ```scanner.py``` is used and maybe some changes to ```enqueue_task``` in ```utils.py```, so I decided against it for now. I think it could potentially be the better option, but I thought I'd just do something a little simpler with the Singleton.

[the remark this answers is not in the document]
korydraughn: Sounds reasonable. Just trying to think of issues that can pop up due to singletons.

From what you've written, Minio is thread-safe, so that's good. Unless someone disagrees, I think you keep rolling with your current approach and continue to test to make sure everything is in working order.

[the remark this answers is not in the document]
alanking: > This means that multiple celery workers could potentially be accessing the instance of the Minio client object at the same time when they each call the upload_file_from_S3 function for a file in their batch of files, which might be an issue?

The documentation is wildly ambiguous about this, but I'm pretty sure that Celery workers spawn processes, not threads. I don't think that they will be using the same Minio client object as another child process to the Celery worker. In any case, creating a single Minio client instance per worker seems like a good idea instead of creating a new one for every path of a given task (by default, it's creating 50 separate Minio clients per chunk per worker!). So I agree with your assessment that this will provide a performance enhancement, even if a marginal one.

> Originally, my plan was to figure out a way to utilize the Scanner object created in the sync_path task in sync_task.py, which already creates a Minio client to get paths to the objects in the to-ingest S3 bucket. I wanted to pass this object beyond the scope of the sync_path task, but it seems like it will require some changes to how the scanner object in scanner.py is used and maybe some changes to enqueue_task in utils.py, so I decided against it for now. I think it could potentially be the better option, but I thought I'd just do something a little simpler with the Singleton.

I think integrating this with the existing Minio client is a fine idea. We should create an issue to investigate this in the future. Maybe you will be the one to do it, but let's make sure it's written down somewhere other than this comment thread. :)

avkrishnamurthy:
Added changes to have scanner do the uploading/syncing of objects. I tested it with register, put, put_sync, and put_append for both S3 files and non-S3 files, parallelized and non-parallelized, and it worked. If this is too big of a change for this PR/issue, I understand and can create a separate issue and PR for refactoring the scanner.py sync_task.py, and sync_irods.py if needed.

irods_capability_automated_ingest/sync_irods.py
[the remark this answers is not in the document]
avkrishnamurthy: New update for this is that by refactoring sync_task.py and having the scanner object to do the upload/syncing, there is no more need to use a Singleton for the Minio client.

irods_capability_automated_ingest/scanner.py
```diff
@@ -216,6 +257,165 @@ def construct_path(self, meta, path):
...
+        bucket_name = path_list[0]
+        object_name = path_list[1]
+
+        MAXIMUM_SINGLE_THREADED_TRANSFER_SIZE = 32 * ( 1024 ** 2)
+
+        total_bytes = self.client.stat_object(bucket_name, object_name).size
+        #Parallelization is only viable for new files, OR if op=PUT_SYNC for existing files since we copy the entire file again
+        if total_bytes > MAXIMUM_SINGLE_THREADED_TRANSFER_SIZE and (not exists or op==Operation.PUT_SYNC):
+            target_resc_name = options.get(kw.RESC_NAME_KW, '') or options.get(kw.DEST_RESC_NAME_KW, '')
+            logger.info("Parallel put", task="irods_S3upload_file")
```
alanking, line 293: Consider adding the path to the log message here

irods_capability_automated_ingest/scanner.py
```diff
@@ -216,6 +257,165 @@ def construct_path(self, meta, path):
...
+            bytecounts = self.parallel_upload_from_S3(logger, session, bucket_name, object_name, dest_dataobj_logical_fullpath, total_bytes, target_resc_name)
+            if bytecounts!=total_bytes:
+                raise RuntimeError("Parallel put from S3 failed: [{}]".format(source_physical_fullpath))
+            
+        #Single stream should happen for files <= 32MiB or for op=PUT_APPEND
+        else:
+            
+            #Ensures tsize is not None
+            tsize = size(session, dest_dataobj_logical_fullpath) or 0
+            logger.info("Single stream put", task="irods_S3upload_file")
```
alanking, line 303: Consider adding the path to the log message here

irods_capability_automated_ingest/scanner.py
```diff
@@ -231,5 +431,34 @@ def scanner_factory(meta):
         return s3_scanner(meta)
     return filesystem_scanner(meta)
 
+def thread_copy_chunk_from_S3(dst_data_object, chunk_range, minio_client, bucket_name, object_name, dst_chunk_mgr, thread_debug_id):
```
alanking, line 434: Can we move this inside the S3 scanner class?

irods_capability_automated_ingest/sync_irods.py
```diff
@@ -12,9 +14,10 @@
 import irods.keywords as kw
 import base64
 import hashlib
+import multiprocessing
 import ssl
 import threading
-
+import concurrent.futures
```
alanking, line 20: Let's revisit these imports to make sure they're all needed now that we've removed some code.

irods_capability_automated_ingest/scanner.py
```diff
@@ -216,6 +257,165 @@ def construct_path(self, meta, path):
...
+            
+        #Single stream should happen for files <= 32MiB or for op=PUT_APPEND
+        else:
+            
+            #Ensures tsize is not None
+            tsize = size(session, dest_dataobj_logical_fullpath) or 0
+            logger.info("Single stream put", task="irods_S3upload_file")
+            self.single_stream_upload_from_S3(logger, session, meta, bucket_name, object_name, dest_dataobj_logical_fullpath, offset=tsize, **options)
+    
+    def parallel_upload_from_S3(self, logger, session, bucket_name, object_name, dest, total_bytes, target_resc_name=''):
```
alanking, line 306: Should we have the kwargs (i.e. `**options`) for this function? Might be missing some functionality here if we don't pass that along.

We're also missing `meta`, but I'm assuming that's because we don't end up using it?

irods_capability_automated_ingest/scanner.py
```diff
@@ -216,6 +257,165 @@ def construct_path(self, meta, path):
...
+        object_name = path_list[1]
+
+        MAXIMUM_SINGLE_THREADED_TRANSFER_SIZE = 32 * ( 1024 ** 2)
+
+        total_bytes = self.client.stat_object(bucket_name, object_name).size
+        #Parallelization is only viable for new files, OR if op=PUT_SYNC for existing files since we copy the entire file again
+        if total_bytes > MAXIMUM_SINGLE_THREADED_TRANSFER_SIZE and (not exists or op==Operation.PUT_SYNC):
+            target_resc_name = options.get(kw.RESC_NAME_KW, '') or options.get(kw.DEST_RESC_NAME_KW, '')
+            logger.info("Parallel put", task="irods_S3upload_file")
+            bytecounts = self.parallel_upload_from_S3(logger, session, bucket_name, object_name, dest_dataobj_logical_fullpath, total_bytes, target_resc_name)
```
alanking, line 294: I feel like splitting up `bucket_name` and `object_name` for the args here is a little awkward. Consider passing `meta` and pulling out the `path` like we do elsewhere. Alternatively, `meta` could get the bucket name and object name added to it.

Acknowledging here that the Minio client takes the bucket and object names separately, which seems to be the driver behind this decision. If this seems okay to everybody else, then I'm fine with how it is.

alanking (commented):
We had some discussion offline... I have some reservations around the dependency entanglement between the scanners and the irods_sync functions, but we can address this at a later time.

Given that this is new functionality, what you have here seems good as long as the S3 uploads are working

irods_capability_automated_ingest/scanner.py
trel: obviously bucket and object are s3-specific things - so doesn't feel wrong to split them by name here to me.    can always use more comments for 'why' things are done.

irods_capability_automated_ingest/sync_irods.py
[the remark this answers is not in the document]
alanking: I think that the new approach is closer to what we should do. This can be resolved!

irods_capability_automated_ingest/scanner.py
avkrishnamurthy: What if instead pass the ```source_physical_fullpath``` (like how dest_dataobj_logical_fullpath is passed), and instead get the ```bucket_name``` and ```object name``` in ```single_stream_upload_from_S3``` and ```parallel_upload_from_S3```? Not sure if that is the best of both worlds but it is another option.

irods_capability_automated_ingest/scanner.py
trel: agreed, kwargs for consistency and flexibility.

Passing the meta feels more consistent - and pulling out what we need inside seems fine.      The upside here is also consistency and future flexibility.   Is there any reason *not* to pass the meta?

alanking: Oh, got it. This is mimicking the parallel upload from PRC. I feel like we should pass it along just in case because other options may be set that we don't know about that need to make their way to the API call...

As for `meta`, I'm fine with whatever makes sense to everybody. I just know that it has been used for pretty much everything, historically.

irods_capability_automated_ingest/scanner.py
alanking: Nah, let's just leave them separated. It will prevent multiple path splits anyway.

irods_capability_automated_ingest/scanner.py
avkrishnamurthy: No reason to not pass meta. Will include both kwargs and meta.

alanking (commented):
I think this looks good to go. Any thoughts, @d-w-moore?

d-w-moore:
will look at today.

irods_capability_automated_ingest/scanner.py
```diff
@@ -134,24 +141,58 @@ def itr(self, meta, obj, obj_stats):
...
+                while True:
+                    buf = sfd.read(BUFFER_SIZE)
+                    if buf == b"":
+                        break
+                    tfd.write(buf)
+            tfd.close()
+            logger.info("succeeded", task="irods_FSappend_file", path=source_physical_fullpath)
+
+        #If file doesn't exist, just do a standard put
+        #If file does exists, we know op=PUT_SYNC, and we re-copy whole file again, so it is fine also
```
d-w-moore, line 180: "file" -> "data object"

d-w-moore (changes requested):
This looks good to me. Nice adaptation of PRC code to the purpose at hand.

korydraughn:
Excellent work!

Issue #222, opened 2024-05-01
Use %-formatting instead of f-strings in `logging` calls


alanking:
BTW with logger functions it's best practice to use % formatting instead of f-strings. See https://pylint.pycqa.org/en/v3.1.0/user_guide/messages/warning/logging-fstring-interpolation.html

_Originally posted by @rmoreas in https://github.com/irods/irods_capability_automated_ingest/pull/221#discussion_r1586534430_

Issue #219, opened 2024-04-30
Breaking change in event handler interface caused by #207


rmoreas:
Adding the `scanner` and `op` as positional arguments in call to `event_handler.call()` in [`sync_irods.py` at line 535](https://github.com/irods/irods_capability_automated_ingest/blob/94b0efb478d922b5e1e65144d269904f795834a7/irods_capability_automated_ingest/sync_irods.py#L535) (see fragment below), breaks the interface of the event handler methods. These arguments should be passed as named/keyword arguments instead.

```python
        # ...
        elif content:
            if put:
                sync = op in [Operation.PUT_SYNC, Operation.PUT_APPEND]
                if sync:
                    event_handler.call(
                        "on_data_obj_modify",
                        logger,
                        sync_file,
                        logger,
                        session,
                        meta,
                        scanner,
                        op,
                        **options
                    )
```

The signature of event handler PEP methods are (e.g. `post_data_obj_modify`):

```python
def post_data_obj_modify(cls, hdlr_mod, logger, session, meta, **options)
```

Because of the above change in #207, the signature of the event handler methods need to change to:

```python
def post_data_obj_modify(cls, hdlr_mod, logger, session, meta, *args, **options)
```

In my opinion, adding `*args` makes the interface unclear. I think that all extra arguments should be passed as named arguments.

alanking:
Good catch. We will think about how best to handle this. Seeing as the ingest tool is pre-1.0 and the interfaces are still (yes, still) settling, we didn't give much consideration for existing deployments. We will write back with our plan after UGM (and possibly after a discussion with the kind folks of KU Leuven!)

rmoreas:
Not directly related directly related, but would be nice that `content` flag passed to `sync_data_from_file` is also passed to the `pre_data_obj_modify` and `post_data_obj_modify` handler methods to be able to differentiate between a data content change or metadata-only change.

alanking:
To restore the interface to match the released one, we decided that we can put the new arguments in the `meta` dict. In this way, the interface only changes in the sense that some new keys appear in the dict. Besides the explicit arguments that currently exist, all the other information passed around internally and through the event handler interface is stored in `meta`, so this seemed the least disruptive.

alanking:
[...]

PR #236, opened 2024-06-27
Restore event handler interface, fix log formatting technique

alanking:
Addresses #219 
Addresses #222 

Tests are passing. Need to test with S3 scanner as well since there are no S3 tests in the regular suite yet.

### DIFF
--- a/irods_capability_automated_ingest/sync_task.py
+++ b/irods_capability_automated_ingest/sync_task.py
@@ -207,7 +207,7 @@ def restart(meta):
 
 @app.task(bind=True, base=IrodsTask)
 def sync_path(self, meta):
-    syncer = scanner.scanner_factory(meta)
+    scanner_instance = scanner.scanner_factory(meta)
     path = meta["path"]
     config = meta["config"]
     logging_config = config["log"]
@@ -225,7 +225,7 @@ def sync_path(self, meta):
         meta["task"] = "sync_dir"
         chunk = {}
 
-        itr = syncer.instantiate(meta)
+        itr = scanner_instance.instantiate(meta)
 
         if meta["profile"]:
             profile_log = config.get("profile")
@@ -253,7 +253,7 @@ def sync_path(self, meta):
             obj_stats = {}
 
             try:
-                full_path, obj_stats = syncer.itr(meta, obj, obj_stats)
+                full_path, obj_stats = scanner_instance.itr(meta, obj, obj_stats)
             except ContinueException:
                 continue
 
@@ -284,8 +284,8 @@ def sync_path(self, meta):
 
 
 def sync_entry(self, meta, cls, datafunc, metafunc):
-    syncer = meta.get('scanner')
-    if syncer is None:
+    scanner_instance = meta.get('scanner')
+    if scanner_instance is None:
         raise RuntimeError("'scanner' not found in options. Cannot sync data.")
 
     path = meta["path"]
@@ -363,7 +363,7 @@ def sync_entry(self, meta, cls, datafunc, metafunc):
                 else:
                     target2 = target
             else:
-                target2 = syncer.construct_path(meta2, path)
+                target2 = scanner_instance.construct_path(meta2, path)
 
             # If the event handler has a character_map function, it should have returned a
             # structure (either a dict or a list/tuple of key-value tuples) to be used for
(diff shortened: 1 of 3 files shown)


korydraughn (commented):
Changes seem fine.

Is there anything else left to do for this?

irods_capability_automated_ingest/sync_task.py
```diff
@@ -283,7 +283,11 @@ def sync_path(self, meta):
         raise self.retry(max_retries=max_retries, exc=err, countdown=retry_countdown)
 
 
-def sync_entry(self, meta, cls, syncer, datafunc, metafunc):
+def sync_entry(self, meta, cls, datafunc, metafunc):
+    syncer = meta.get('scanner')
```
trel, line 287: just pointing out that these are not consistent.    scanner vs syncer.

trel:
if we're using meta for 'scanner'... do we want to use that name everywhere for consistency?

or do we need a better name that encapsulates its various (currently appears to be 'two' different) uses?

alanking:
> if we're using meta for 'scanner'... do we want to use that name everywhere for consistency?
> 
> or do we need a better name that encapsulates its various (currently appears to be 'two' different) uses?

The name of the variable does not *need* to match the name of the key in the `meta` dict here, but I suppose I agree that it would make things more consistent. I was just matching what was already there in order to minimize the diff, but can update it. What are the "two" uses?

trel:
the two i see are 'scanner' and 'syncer'.  is there a better 'composite' word?   or is 'scanner' sufficient?

alanking:
I see. The only thing I would be concerned about at this point is the key being used in the `meta` map because that could be accessed by a custom event handler - although that seems extremely inadvisable :) - and changing that in the future would be another change to the interface. The name of the variable is less important because it is internal to each function (or at least it is now).

The term "scanner" was chosen to match the name of the class, and the name of the class was chosen, IIRC, to reflect the fact that the class is responsible for implementing a way of walking or "scanning" different storage systems. This is in reference to the filesystem scanner use case as well, but there are of course other use cases that we advertise (namely, landing zone).

I think "syncer" was used due to the filesystem "sync" referenced throughout this codebase and in our training. We have operations that are not considered "sync"s (e.g. `PUT_APPEND`) so not everything being done is a sync. So, I would be more in favor of changing "syncer" to "scanner", but I'm not attached to that term, really. Open to suggestions.

trel:
> So, I would be more in favor of changing "syncer" to "scanner"

I think that is correct and good.  Let's make that change and see how it looks/feels.

alanking:
In order to avoid ambiguity (i.e. syntax errors, and... confusion) can we differentiate the `scanner` class and instantiations by name?

Here's what can happen if we name the instance the same thing as the class:
```python
>>> class scanner:
...     def __init__(self, x):
...             self.x = x
...     def foo(self):
...             self.x = self.x + 1
... 
>>> scanner = scanner(1)
>>> scanner.foo()
>>> scanner.x
2
>>> scanner.foo()
>>> scanner.x
3
>>> scanner2 = scanner(10)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'scanner' object is not callable
```
Once the `scanner` is instantiated as `scanner`, you cannot instantiate any more `scanner`s because the class `scanner` has been shadowed by the instance named `scanner`.

With that in mind, here are some suggestions...
 - scanner_instance
 - storage_scanner
 - path_scanner
 - walker / crawler
 - syncer (yes, I know, just... consider it in this new light)

alanking:
We could also consider renaming the class, but I think we can / should do that later...

trel:
storage_scanner seems better than path_scanner b/c S3 isn't 'paths' officially.

walker and crawler are nice... the scanner is the technology/approach, and the walker/crawler is the thing doing the work.

trel:
although upon 30 more seconds of thought... i like `scanner_instance`.

that way we don't invent new language/nouns... but avoid the collision.

irods_capability_automated_ingest/sync_task.py
alanking: I've renamed all instances of `syncer` with `scanner_instance`, per the discussion in the main thread for this PR. Resolving.

trel: nice.

trel:
looking real good.

korydraughn:
Agreed. `scanner_instance` is a good choice.

alanking:
Confirmed that a basic S3 bucket scan worked with these changes.

korydraughn (approved):
Squash to taste and pound it.

alanking:
I like the current commits, actually, so just added #s. Mergin

PR #237, opened 2024-07-01
Add demo Compose project for testing

alanking:
No issue number yet...

Derived from the https://github.com/irods/irods_demo Compose project.

I used this to confirm S3 scans (no automated tests yet). Not sure if we want to include this in the project but I'm putting it up just in case anyway.


docker/demo/ingest_celery_workers/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.11
+
+ARG PIP_PACKAGE="irods-capability-automated-ingest"
+
+RUN pip install ${PIP_PACKAGE}
+
+ENTRYPOINT ["celery", "-A", "irods_capability_automated_ingest.sync_task", "worker", "-Q", "restart,path,file"]
+#CMD ["-h"]
```
korydraughn, line 8: Looks like this can be removed.

docker/demo/compose.yaml
```diff
@@ -0,0 +1,63 @@
...
+            redis:
+                condition: service_started
+            irods-catalog-provider:
+                condition: service_healthy
+
+    minio:
+        image: minio/minio:RELEASE.2024-06-28T09-06-49Z
+        command: server --address ":19000" --console-address ":19001" /data/minio-s3
+        ports:
+            - "19000:19000"
```
korydraughn, line 53: What is port 19000 for again?

Consider leaving a comment for that so people unfamiliar with the project don't have to guess/investigate.

docker/demo/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+```
+docker compose up
+```
+
+## Stop
+
+```
+docker compose down
+```
```
korydraughn, line 37: Are more words planned for these sections?

docker/demo/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Scanning an S3 bucket
+
+Change the port exposed by the `minio` service, if needed, so that the MinIO Console can be accessed. The MinIO server is being run by the user `irods` and `irodsadmin`. The place from which the job is launched should have a keypair file with these credentials:
```
korydraughn, line 41: The part `and irodsadmin` in the following reads a little weird.
```
The MinIO server is being run by the user `irods` and `irodsadmin`.
```
I take it those are the example credentials. Consider rewording so that it's clear what those are.

korydraughn (commented):
Seems correct.

docker/demo/README.md
alanking: Oh, yeah, that's pretty bad. I kinda threw this together right before opening the PR just so I wouldn't forget how to use this. :p I'll rephrase that.

docker/demo/README.md
alanking: I don't think so. These are more or less placeholder sections in case more words are needed later. There aren't really any runtime options at the moment, and stopping the containers isn't particularly special either.

docker/demo/ingest_celery_workers/Dockerfile
alanking: Yeap

docker/demo/compose.yaml
alanking: I'll leave a comment in the file. 19000 is the default port for issuing S3 requests. In other words, this is the port to use when targeting the S3 endpoint for a bucket scan.

docker/demo/README.md
alanking: Done

docker/demo/compose.yaml
alanking: Done

docker/demo/ingest_celery_workers/Dockerfile
alanking: Done

docker/demo/README.md
korydraughn: Gotcha. When I reached those sections, it wasn't clear to me what to expect.
- Are tests supposed to run automatically on `docker compose up`?
    - _That is the current behavior of the S3 API test harness._
- Or do the commands just stand things up and nothing else?

alanking: Oh, there are no tests as of yet. This just runs the iRODS server, the Celery workers, a Redis database, and a MinIO server and it just sits there.

korydraughn: Ok. That aligns with what I thought it did.

Consider adding a line stating that unless there are plans to change those sections.

alanking: I reworked the section to include that description and some other things. Resolving this one, but the conversation can continue based on the new changes.

korydraughn (commented):
Looks good. We'll have a better idea of what to do with this once we figure out the S3 testing situation.

alanking:
I have found this addition to be very helpful in quickly iterating for some new features. One can copy files into the container and restart the ingest worker service to apply changes very easily, and this works much better than the existing Docker stuff in this repo.

alanking:
Possibly. It may (or may not) be better to use volume mounts to install the package rather than building it into the image as is being done now. This would allow the developer to make changes in the host machine and changes would be applied to all instances of the ingest worker service rather than having to copy the files into each container. This would of course be an alternative approach, and may have some drawbacks from the current implementation. I don't think we should go for that right away.

Anyways, we can add a section to the README with a "suggested workflow" for developers. As is implied above, there are probably better ways to do this, so I'm a little hesitant to say how one should use this thing.

korydraughn:
That's reasonable. We can always add words to the README or somewhere else if it proves to be helpful.

What's left for this PR?

alanking:
I think I need to adjust the default number fed to the `-c` option for Celery. Using the default - which matches the number of cores on the host - causes the iRODS server to become unusable for modestly large ingest jobs, so it would be better to reduce this to, like, 2 or even 1. This would probably involve adding a `CMD` line which can be overridden in the Compose YAML file or using the command line options.

docker/demo/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+It's easiest to try out scanning things from the `ingest-celery-worker` service instance.
+
+## Build
+
+```
+docker compose build
+```
+
+The ingest-celery-worker service has a build argument that allows for controlling the version of the ingest package. Here's how to use it:
```
trel, line 15: please note what happens if `--build-arg` is not specified.

docker/demo/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Scanning an S3 bucket
+
+Change the port exposed by the `minio` service, if needed, so that the MinIO Console can be accessed. The MinIO server is being run with access key `irods` and secret key `irodsadmin`. The place from which the job is launched should have a keypair file with these credentials:
+```
+irods
+irodsadmin
+```
+
+To perform a basic scan of an S3 bucket called, for example, `ingest-test-bucket`, run something like the following:
```
trel, line 55: please note this is run 'inside' a specific container.

docker/demo/compose.yaml
```diff
@@ -0,0 +1,63 @@
+version: '3'
```
trel, line 1: i think this is no longer needed

docker/demo/irods_catalog_provider/setup.input
```diff
@@ -0,0 +1,28 @@
+
```
trel, line 1: would love for setup to use unattended install (`--json_configuration_file`).

alanking: Can I plead lazy and do that once this issue is done? :) https://github.com/irods/irods_demo/issues/65

trel: but of course

trel: same in irods_demo

docker/demo/compose.yaml
alanking: Will do. I think I should specify a different default project name here, too, since "demo" makes for some opaque image names:
```
demo-ingest-celery-workers
demo-irods-catalog-provider
demo-irods-catalog
```

trel: yep.   `irods-ingest` might be good enough for a prefix.

alanking: I put `irods-ingest-demo` but I'm willing to change it.

docker/demo/README.md
alanking: Done

docker/demo/README.md
alanking: Done

docker/demo/README.md
```diff
@@ -24,6 +24,11 @@ This will clone the specified git repository and checkout the commit-ish specifi
 docker compose build --build-arg PIP_PACKAGE=irods-capability-automated-ingest==0.4.2
 ```
 
+If no `--build-arg` is specified, the default build will install the latest released version of the package from PyPi. The following is equivalent to not specifying a `--build-arg` when building the project:
```
trel, line 27:
```suggestion
If no `--build-arg` is specified, the default build will install the latest released version of the package from PyPI. The following is equivalent to not specifying a `--build-arg` when building the project:
```

docker/demo/compose.yaml
trel: sold

docker/demo/README.md
alanking: Done

alanking:
Well... I suppose I can mark this as ready for review now

trel:
ha.   i think squash em.   certainly good enough to get in - no functional changes to the software directly.

alanking:
Had one last thing regarding the Celery concurrency adjustments. I tried doing this with command line options but Compose does some weird stuff (creates containers with a `-run-<sha>` suffix) when using `docker compose run` and the like, so it's probably just best to make adjustments before the containers are created for the service instances. A note saying as much has been added to the README and the YAML file.

docker/demo/README.md
```diff
@@ -0,0 +1,72 @@
...
+Change the port exposed by the `minio` service, if needed, so that the MinIO Console can be accessed. The MinIO server is being run with access key `irods` and secret key `irodsadmin`. The place from which the job is launched should have a keypair file with these credentials:
+```
+irods
+irodsadmin
+```
+
+To perform a basic scan of an S3 bucket called, for example, `ingest-test-bucket`, run something like the following:
+
+```
+python3 -m irods_capability_automated_ingest.irods_sync start /ingest-test-bucket /tempZone/home/rods/ingest-test-bucket --s3_keypair /path/to/s3keypair.txt --s3_endpoint_domain minio:19000 --s3_insecure_connection  --synchronous --progress
```
korydraughn, line 69: Consider breaking this line across multiple lines so it's easy for the reader to see what's happening. For example:
```
python3 -m irods_capability_automated_ingest.irods_sync start \
    /ingest-test-bucket \
    /tempZone/home/rods/ingest-test-bucket \
    --s3_keypair /path/to/s3keypair.txt \
    --s3_endpoint_domain minio:19000 \
    --s3_insecure_connection \
    --synchronous \
    --progress
```

alanking:
Okay, squashed and given issue #244.

docker/demo/ingest_celery_workers/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.11
+
+ARG PIP_PACKAGE="irods-capability-automated-ingest"
```
korydraughn, line 3: Hmm, should we add `IRODS_` to the arg name?

I'm leaning towards yes. That would be a clear indicator that it's an iRODS thing rather than a pip thing.

Thoughts?

docker/demo/README.md
alanking: Done

docker/demo/irods_catalog_provider/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+        irods-runtime \
+        irods-server \
+    && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/*
+
+COPY setup.input /
+RUN mv /setup.input /irods_setup.input
+
+WORKDIR /
```
korydraughn, line 44: Isn't this the default?

This isn't a request to change anything. Just wanted to be sure of the "why".

docker/demo/ingest_celery_workers/Dockerfile
alanking: I think `PIP_PACKAGE` is fine as it is referring to the application being run by this image (i.e. it only does one thing). If we are going to change this, we should be even more specific; perhaps `IRODS_AUTOMATED_INGEST_PIP_PACKAGE` or `IRODS_INGEST_PIP_PACKAGE`. This is not a pip package for iRODS, so `IRODS_PIP_PACKAGE` seems worse to me.

docker/demo/irods_catalog_provider/Dockerfile
alanking: This was copied from `irods_demo`. I don't know why it's there (despite me being the one that added it there as well). Perhaps to avoid getting bitten by a change of the default in the future?

korydraughn: Quite possibly. Either way, it's fine.

Consider removing it since it doesn't affect the outcome of the image.

alanking: I removed the `WORKDIR` and made all following paths absolute to be sure. Seems to be working. Resolving.

docker/demo/ingest_celery_workers/Dockerfile
korydraughn: The overall name doesn't concern me. `PIP_PACKAGE` seems like it could be a real option supported by pip itself.

Any of the more explicit variable names you proposed satisfy my concern.

> I think `PIP_PACKAGE` is fine as it is referring to the application being run by this image (i.e. it only does one thing).

That's a fair point. The parameter is only used as a build argument (i.e. it's never defined as an environment variable, etc), so it can't necessary collide with anything or introduce weirdness during the build step.

alanking: Okay. I went with `IRODS_AUTOMATED_INGEST_PIP_PACKAGE`. Seems to be working. Resolving.

korydraughn:
I think we're ready for squashing.

alanking:
Squashed

korydraughn (approved):
If you're happy with it, let's pound it and get it merged.

alanking:
#'d. Mergin

Issue #240, opened 2024-07-12
`pre_coll_modify`/`post_coll_modify` invoked on first sync


alanking:
When performing a sync for the very first time to a collection which does not exist, I would expect `pre_coll_create` and `post_coll_create` to be invoked and any syncs to the same collection after that would be considered updates and invoke `pre_coll_modify` and `post_coll_modify`. However, it seems that `pre_coll_modify` and `post_coll_modify` are invoked on the first sync as well.

I'm wondering whether this is the expected behavior. I did not expect it, so I am labeling it as a bug. I'm willing to be convinced otherwise, in which case, perhaps this can be closed as invalid.

trel:
agreed, feels like a bug on first read.

rmoreas:
Just an idea, I didn't check the code: maybe that the modified is triggered because of dataobjects/subcollections are added after the collection has been created?

korydraughn:
This seems relevant. @alanking Can you derive anything from this code that may explain the behavior.

https://github.com/irods/irods_capability_automated_ingest/blob/76dced0c684156720f8064790de4399ad8cbff7c/irods_capability_automated_ingest/sync_irods.py#L574-L599

This is the only other place I see referencing `on_coll_create`.

https://github.com/irods/irods_capability_automated_ingest/blob/76dced0c684156720f8064790de4399ad8cbff7c/irods_capability_automated_ingest/sync_irods.py#L46-L73

alanking:
> Just an idea, I didn't check the code: maybe that the modified is triggered because of dataobjects/subcollections are added after the collection has been created?

Rationally, I think that I agree with this justification of the observed behavior. I mostly filed this issue in order to discuss whether this is the expected behavior, but perhaps "desired behavior" would have been a better term. After thinking about it some more, I think that this is in fact the desired behavior. The creation of a collection is a completely separate task from adding data objects and sub-collections, and this is required in order to keep everything fully parallel. So, it follows that adding a data object or sub-collection to a collection can be said to be "modifying" the collection.

However, now I'm concerned about the "how" because I do not see where `coll_modify` can be triggered outside of a few key places. Let me see if I can map this out coherently...

[...]

PR #241, opened 2024-07-12
Add *args to all example event handler methods, and add tests for pre/post event handler methods

alanking:
Addresses #219 (for real)

This partially reverts the work done in https://github.com/irods/irods_capability_automated_ingest/pull/236.

This change primarily adds tests for almost all relevant operations for all pre/post event handler methods that we currently support. This is meant to act as a minimal baseline to ensure that the pre/post methods are executing properly.

Case in point: Possible bug was discovered while writing these tests and has been created here: #240.

All new tests pass. Running full suite now.

### DIFF
--- a/irods_capability_automated_ingest/test/test_irods_sync.py
+++ b/irods_capability_automated_ingest/test/test_irods_sync.py
@@ -23,6 +23,7 @@ from os.path import (
     isfile,
 )
 from irods.session import iRODSSession
+from irods.meta import iRODSMeta
 from irods.models import Collection, DataObject
 from tempfile import NamedTemporaryFile, mkdtemp
 from datetime import datetime
@@ -34,6 +35,7 @@ from irods_capability_automated_ingest.sync_job import sync_job
 from irods.data_object import irods_dirname, irods_basename
 import irods_capability_automated_ingest.examples
 import irods.keywords as kw
+from irods_capability_automated_ingest.utils import Operation
 
 os.environ["CELERY_BROKER_URL"] = "redis://redis:6379/0"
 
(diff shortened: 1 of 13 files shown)


irods_capability_automated_ingest/test/test_irods_sync.py
```diff
@@ -2290,6 +2292,408 @@ class Test_irods_sync_UnicodeEncodeError(
...
+                        # Run the sync job and ensure that no failures occurred.
+                        self.run_sync(destination_collection, temporary_event_handler.name, job_name)
+                        self.assertEqual(sync_job(job_name, get_redis()).failures_handle().get_value(), None)
+
+                        # Ensure that no metadata items have been annotated because the collection was not modified.
+                        #self.assertEqual(len(session.collections.get(destination_collection).metadata.items()), 0)
+
+                        # TODO(#240): coll_modify is called after coll_create or after each chunk of files is created
+                        # and I would not have expected this to happen until something was actually updated.
+                        self.assertEqual(len(session.collections.get(destination_collection).metadata.items()), 2)
```
alanking, line 2660: I'm not sure if this is the correct behavior. The assertion shown here is true. The commented-out one is false.

alanking (commented):
May need to rename `syncer` to `scanner_instance` again if we want to do that. Seems less important since we are not using the "scanner" key in `meta` anymore, though. Up to reviewers.

irods_capability_automated_ingest/test/test_irods_sync.py
alanking: Depending on how we answer this question (see/discuss in #240) that would also impact our list of expected AVUs.

trel:
this is really nice.

i do still think scanner_instance is better than syncer.

korydraughn:
Yeah, scanner_instance is better IMO.

korydraughn (commented):
Changes look good.

What's remaining for this?

alanking:
> What's remaining for this?

I was waiting on the full test suite to finish running, which it has. Everything is passing.

After that, I was going to see if #240 is a real issue or not. If so, I need to update my tests and fix it. If not, I need to remove the TODO from my code. If we want to punt on that question til after 0.5.0, I can leave things as they are.

Also, I can go ahead and do the `syncer` -> `scanner_instance rename` again.

korydraughn:
Given the behavior described in #240 has always been that way, I don't think it's a blocker for 0.5.0.

I say let's punt and get 0.5.0 released _(after renaming `syncer`)_.

irods_capability_automated_ingest/test/test_irods_sync.py
alanking: It has been decided that we will address this issue at a later time: https://github.com/irods/irods_capability_automated_ingest/pull/241#issuecomment-2228861748

I am going to leave this as is. Resolving.

alanking:
Okay. I bumped #240 out of 0.5.0, resolved the open conversation, and renamed all instances of `syncer` to `scanner_instance`.

korydraughn (commented):
Squash to taste.

alanking:
#'d. Mergin.

Issue #289, opened 2024-10-02
`Operation.PUT` does not error when attempting to update existing data objects


alanking:
The README update made in https://github.com/irods/irods_capability_automated_ingest/commit/c6e0f266ff613db8d47575f9376e63bb6a0da476 erroneously states that `Operation.PUT` will emit an error if the data object being synced exists. This is not the case and has not been the case for as long as the operation has existed. Please update the documentation to reflect what actually happens, as stated in the previous documentation: the operation is ignored and no data is synced.

PR #290, opened 2024-10-02
Fix PUT operation documentation and give node name to demo Celery workers

alanking:
Addresses #289

The node name thing is to prevent a warning message that appears if multiple Celery worker nodes are running in the same host.


korydraughn (approved):
Pound and merge it.

alanking:
Fixed commit message, #'d, mergin

Issue #293, opened 2024-10-03
Scheduling a periodic job fails on main


alanking:
The renamed main task from `restart` to `filesystem_main_task` and `s3_bucket_main_task` causes these lines to fail:
https://github.com/irods/irods_capability_automated_ingest/blob/b28e5c486c5476c05f85d44ece5aed5f40f3492f/irods_capability_automated_ingest/tasks/filesystem_tasks.py#L109-L112
https://github.com/irods/irods_capability_automated_ingest/blob/b28e5c486c5476c05f85d44ece5aed5f40f3492f/irods_capability_automated_ingest/tasks/s3_bucket_tasks.py#L30-L33

Issue #210, opened 2023-07-21
Synchronous start job command doesn't exit when the job is stopped with the cli


rmoreas:
When a synchronous sync job is started, the command doesn't exit when the corresponding job is stopped thought the CLI. The start job command keeps waiting for `job.done()` forever.

Because of this issue it's not possible to interrupt a synchronous sync job started from a cronjob. The cronjob keeps running forever, unless the process is killed.

The root cause seems to be that the [`sync_job.done`](https://github.com/irods/irods_capability_automated_ingest/blob/main/irods_capability_automated_ingest/sync_job.py#L40) function checks the `tasks` key of the job, but this key is not removed by the [`stop_job`](https://github.com/irods/irods_capability_automated_ingest/blob/main/irods_capability_automated_ingest/sync_actions.py#L17) action.

rmoreas:
Maybe add `job.reset()` after [line 27 of `sync_actions.py`](https://github.com/irods/irods_capability_automated_ingest/blob/main/irods_capability_automated_ingest/sync_actions.py#L27)?

rmoreas:
Hi Alan, could you have a quick look at this one? What do your think about the proposed solution?

alanking:
Sorry for the delay...

At a glance, I think the proposed solution sounds reasonable. I would need to do some more reading about `interrupt` but calling `cleanup` after stopping a job seems to just make sense.

alanking:
Sorry for the very long delay in this...

I have confirmed that the proposed fix does indeed cause the `start` subcommand to return. There's no indication that the job was stopped in any way - it just exits with an error code of 0. If `--progress` is set, the task count is set to nothing and the progress bar fills up once the job is stopped:
```
 [Elapsed Time: 0:00:08] |##############################| (Time:  0:00:08) count: ------ tasks: ------ failures: ------ retries: ------
```

I think this is better than waiting forever (requiring the script to be terminated).

If we wanted to communicate that the job was stopped, we would need to add another Redis query to see that there are tasks but that the job is gone. I'm not sure how much value that adds.

korydraughn:
> There's no indication that the job was stopped in any way - it just exits with an error code of 0.

Do you know if the task still runs after the subcommand returns control?

alanking:
Then again, we probably want to reset everything else for the job, too, so, calling `reset` is, in the end, the most straightforward approach.

[...]

Issue #93, opened 2019-05-14
watch subcommand should cleanly handle KeyboardInterrupt


trel:

alanking:
Is the idea here to catch the `KeyboardInterrupt` exception and exit with a code of 0?

trel:
I believe so, yes.   Just 'stop' watching and let go cleanly.

Please paste here what happens today so we can confirm.

alanking:
```
$ python3 -m irods_capability_automated_ingest.irods_sync watch a5273a18-81a0-11ef-9921-0242ac190006
 [Elapsed Time: 0:00:03] |####################################################################################################################################                                  | (ETA:   0:00:00) count:      5 tasks:      1 failures: ------ retries: ------^C
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/local/lib/python3.11/site-packages/irods_capability_automated_ingest/irods_sync.py", line 424, in <module>
    main()
  File "/usr/local/lib/python3.11/site-packages/irods_capability_automated_ingest/irods_sync.py", line 420, in main
    sys.exit(args.func(args))
             ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/irods_capability_automated_ingest/irods_sync.py", line 210, in handle_watch
    return sync_actions.monitor_job(args.job_name, True, get_config(args))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/irods_capability_automated_ingest/sync_actions.py", line 112, in monitor_job
    time.sleep(1)
KeyboardInterrupt

$ echo $?
130
```

trel:
Right, okay - yes, please make it cleanly return a 0.   No barfing on the customer.

Issue #92, opened 2019-05-14
watch subcommand should show total runtime/ETA, not calculate from invocation


trel:
When watching/attaching to a job that has been running for sometime, the current status of the count, tasks, failures, and retries are correct... but the runtime and ETA are calculated from the moment that the watch command was issued, not when the job was created.

PR #294, opened 2024-10-04
`irods_sync` script fixes and enhancements (plus one other bug fix)

alanking:
Addresses #91 
Addresses #92 
Addresses #93 
Addresses #210 
Addresses #293

I'm contemplating adding tests for these subcommands, but it seems a little tricky (hence, draft PR) and there are no existing tests for any of the subcommands other than `start`. Existing tests are passing.

Also, the two commits referencing #210 (only) can be squashed into one if we like the solution of maintaining a list of stopped jobs. Otherwise, we can drop the second commit because the first one addresses the issue (albeit, not as well, IMO).

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,69 @@ logger = sync_logging.get_sync_logger(log_config)
 ```
 The returned logger can then be used to log messages in a standard way. Note: This should only be done inside a Celery task or Event Handler method.
 
-## `irods_sync start`
+## `irods_sync` script
+
+### Common options
+
+#### Redis configuration options
+
+These options can be used in lieu of the `CELERY_BROKER_URL` environment variable to point the `irods_sync` script at a particular Redis instance.
+
+##### `--redis_host`
+
+`--redis_host REDIS_HOSTNAME` will use `REDIS_HOSTNAME` as the hostname for the Redis database with which the `irods_sync` script will communicate. The default `REDIS_HOSTNAME` used in the absence of any other input is `localhost`.
+
+##### `--redis_port`
+
+`--redis_port REDIS_PORT` will use `REDIS_PORT` as the port number on which the Redis database is listening for incoming requests. The default `REDIS_PORT` used in the absence of any other input is 6379, which is the official default Redis port number.
+
+##### `--redis_db`
+
+`--redis_db REDIS_DB` will use `REDIS_DB` as the database number where essential Automated Ingest information is stored. The default `REDIS_DB` used in the absence of any other input is 0, which is the official default Redis database number.
+
+#### Logging options
+
+##### `--log_level`
+
+`--log_level LOG_LEVEL` specifies that only messages of level `LOG_LEVEL` or higher should be logged by the `irods_sync` script. `LOG_LEVEL` must be one of the following string values (listed in ascending order):
+
+ - NOTSET
+ - DEBUG
+ - INFO
+ - WARNING
+ - ERROR
+ - CRITICAL
+
+By default, only messages of level `WARNING` or higher will be logged.
+
+Celery worker log messages are not affected by this option. To specify a log level for the Celery workers, the `--loglevel` option should be used when running the `worker`. See [Celery documentation](https://docs.celeryq.dev/en/stable/reference/cli.html#cmdoption-celery-worker-l) for more information.
+
+##### `--log_filename`
+
+`--log_filename /path/to/log_file.txt` will use `/path/to/log_file.txt` as the destination for log messages emitted by the `irods_sync` script. The default value is `None`, which means that log messages go to `stdout`.
+
+Celery worker log messages are not affected by this option. To specify a log file for the Celery workers, the `--logfile` option should be used when running the `worker`. See [Celery documentation](https://docs.celeryq.dev/en/stable/reference/cli.html#cmdoption-celery-worker-f) for more information.
+
+##### `--log_when`
+
+`--log_when WHEN` will use `WHEN` as the value passed to the `when` parameter of the `TimedRotatingFileHandler` used for managing log files. See [Python `TimedRotatingFileHandler` documentation](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.TimedRotatingFileHandler) for more information. This option only has an effect if `--log_filename` is used.
+
+##### `--log_interval`
+
+`--log_interval INTERVAL` will use `INTERVAL` as the value passed to the `interval` parameter of the `TimedRotatingFileHandler` used for managing log files. See [Python `TimedRotatingFileHandler` documentation](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.TimedRotatingFileHandler) for more information. This option only has an effect if `--log_when` is used. The default value is **1**.
+
+#### Profiling options
+
+`--profile` allows you to use [vis](https://visjs.org) to visualize a profile of Celery workers over time of ingest job.
+
+| option | effect | default |
+| ----   |  ----- |  ----- |
+| profile_filename | Specify name of profile filename (JSON output) | None |
+| profile_level | Minimum level of message to log for profiling | None |
+| profile_interval | Time interval with which to rollover ingest profile file | None |
+| profile_when | Type/units of profile_interval (see TimedRotatingFileHandler) | None |
+
+### `irods_sync start`
 
 `irods_sync start` is used to start a sync job by enqueueing the main task of the overall sync job. The main task is run asynchronously by default, but can be run synchronously with the `--synchronous` option.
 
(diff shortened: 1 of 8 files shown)


irods_capability_automated_ingest/sync_actions.py
```diff
@@ -59,50 +59,46 @@ def monitor_job(job_name, progress, config):
             if failures is not None and failures != 0:
                 return -1
             return 0
+        start_time = job.start_time_handle().get_value()
+        if start_time is None:
```
trel, line 91: why/how could it get in here?

README.md
```diff
@@ -249,7 +249,69 @@ logger = sync_logging.get_sync_logger(log_config)
...
+ - ERROR
+ - CRITICAL
+
+By default, only messages of level `WARNING` or higher will be logged.
+
+Celery worker log messages are not affected by this option. To specify a log level for the Celery workers, the `--loglevel` option should be used when running the `worker`. See [Celery documentation](https://docs.celeryq.dev/en/stable/reference/cli.html#cmdoption-celery-worker-l) for more information.
+
+##### `--log_filename`
+
+`--log_filename /path/to/log_file.txt` will use `/path/to/log_file.txt` as the destination for log messages emitted by the `irods_sync` script. The default value is `None`, which means that log messages go to `stdout` by default.
```
trel, line 291: default ... default (used twice in the same sentence)

README.md
```diff
@@ -249,7 +249,69 @@ logger = sync_logging.get_sync_logger(log_config)
...
+
+Celery worker log messages are not affected by this option. To specify a log file for the Celery workers, the `--logfile` option should be used when running the `worker`. See [Celery documentation](https://docs.celeryq.dev/en/stable/reference/cli.html#cmdoption-celery-worker-f) for more information.
+
+##### `--log_when`
+
+`--log_when WHEN` will use `WHEN` as the value passed to the `when` parameter of the `TimedRotatingFileHandler` used for managing log files. See [Python `TimedRotatingFileHandler` documentation](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.TimedRotatingFileHandler) for more information. This option only has an effect if `--log_filename` is used.
+
+##### `--log_interval`
+
+`--log_interval INTERVAL` will use `INTERVAL` as the value passed to the `interval` parameter of the `TimedRotatingFileHandler` used for managing log files. See [Python `TimedRotatingFileHandler` documentation](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.TimedRotatingFileHandler) for more information. This option only has an effect if `--log_when` is used. The default value is **1**.
```
trel, line 301: why bold the 1, rather than backticks?

README.md
```diff
@@ -249,7 +249,69 @@ logger = sync_logging.get_sync_logger(log_config)
...
+
+#### Profiling options
+
+`--profile` allows you to use vis to visualize a profile of Celery workers over time of ingest job.
```
trel, line 305: is `vis` a thing we need to link to or document?

README.md
alanking: I tend to use backticks for code or command-line options and things of that sort. This is a value for `--log_when`, which I tend to use quotation marks to denote for strings. We don't have to bold it, but I am not a fan of backticks around numeric values. Willing to be convinced.

README.md
alanking: I am actually unsure as to what it is... Maybe this? https://visjs.org/

README.md
korydraughn: This means we need to come up with rules for styling documentation. It's same thing you see at the beginning of technical books.

Or, we let each project use its own styling and just make sure it's consistent throughout the project.

korydraughn:
Are you looking for feedback on this PR even though it's in draft?

alanking:
I want to confirm that this is doing what we want before marking as ready for review. Feel free to leave feedback on what you see, but the implementation may change significantly depending on how the requirements settle out.

korydraughn:
Ok. Will put eyes on this once it's out of draft.

irods_capability_automated_ingest/sync_actions.py
alanking: Is there anything to do for this?

README.md
alanking: Fixed wording

README.md
alanking: Is there anything to do for this?

README.md
alanking: Linked

alanking:
I squashed the two #210 commits together since we're pretty into the stopped jobs list. The stopped jobs list now includes all the information shown in the active jobs list subcommand output. I think this is ready for review...

README.md
korydraughn: The comment about defining rules for documentation doesn't need to happen now.

I'm fine with the current state of things.

korydraughn:
Ok. Will get eyes on it.

irods_capability_automated_ingest/sync_actions.py
trel: no, i think it's fine - just was curious the thinking.. it's basically a null check.

irods_capability_automated_ingest/sync_job.py
```diff
@@ -5,6 +5,24 @@
...
+
+def add_stopped_job(redis_handle, stopped_job_dict):
+    stopped_jobs_handle = redis_key.stopped_jobs_key_handle(redis_handle)
+    stopped_jobs_value = stopped_jobs_handle.get_value()
+    if stopped_jobs_value is None:
+        stopped_jobs_list = []
+    else:
+        stopped_jobs_list = json.loads(stopped_jobs_value.decode("utf-8"))
+    stopped_jobs_list.append(stopped_job_dict)
+    stopped_jobs_handle.set_value(json.dumps(stopped_jobs_list))
```
MartinFlores751, line 18: Do we need to convert to a string here, or could the `set(...)` of the redis handle take care of this?

irods_capability_automated_ingest/sync_job.py
```diff
@@ -5,6 +5,24 @@
...
 import progressbar
+import time
+
+def add_stopped_job(redis_handle, stopped_job_dict):
+    stopped_jobs_handle = redis_key.stopped_jobs_key_handle(redis_handle)
+    stopped_jobs_value = stopped_jobs_handle.get_value()
+    if stopped_jobs_value is None:
+        stopped_jobs_list = []
+    else:
+        stopped_jobs_list = json.loads(stopped_jobs_value.decode("utf-8"))
```
MartinFlores751, line 16: Just to be sure, this always returns a JSON array and never a JSON object at the top level, right? (`[...]` vs `{...}`)

Otherwise, I think line 17 would error out.

irods_capability_automated_ingest/sync_job.py
```diff
@@ -90,3 +109,38 @@ def interrupt(self, cli=True, terminate=True):
...
+        failures = int(self.failures_handle().get_value() or 0)
+        retries = int(self.retries_handle().get_value() or 0)
+        return {
+            "job_name": self.job_name,
+            "total_tasks": total,
+            "remaining_tasks": tasks,
+            "failed_tasks": failures,
+            "retried_tasks": retries,
+            "elapsed_time": time.strftime("%H:%M:%S", time.gmtime(elapsed_time)),
+            "start_time": time.strftime("%Y-%m-%d %H:%M:%S", time.gmtime(start_time)),
```
MartinFlores751, line 145: Is it conveyed anywhere we use `gmtime(...)` vs `localtime(...)`? Is this expected or should we make this more clear with a `%z` for the timezone offset?

MartinFlores751 (commented):
A few comments.

irods_capability_automated_ingest/sync_job.py
alanking: Correct, this is intended to be a JSON array. Maybe I should leave some docstrings on these new things...

irods_capability_automated_ingest/sync_job.py
alanking: Good point... It's possible that redis-py could handle this, but I haven't tried it. The existing descendant of `json_redis_key_handle`, `json_redis_key_handle`, is dumping the JSON to a string before setting the value, so that's why I'm doing it here too.

I think I'd like to at least have the `json_redis_key_handle` do this for us so we don't need to worry about it. I've created https://github.com/irods/irods_capability_automated_ingest/issues/297 for that effort.

irods_capability_automated_ingest/redis_key.py
```diff
@@ -108,6 +106,13 @@ def get_value(self):
         return float(val)
 
 
+# TODO(#292): python metaclasses - see PRC
+# sync_time_key - float with last time particular path was sync'd
+class sync_time_key_handle(float_redis_key_handle):
+    def __init__(self, redis_handle, path):
+        super().__init__(redis_handle, "sync_time", path)
```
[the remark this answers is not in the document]
alanking: Done

irods_capability_automated_ingest/redis_key.py
```diff
@@ -155,3 +160,9 @@ def __init__(self, redis_handle, job_name):
 class retries_key_handle(incremental_redis_key_handle):
     def __init__(self, redis_handle, job_name):
         super().__init__(redis_handle, "retries", job_name)
+
+
+# stopped_jobs_key - value: JSON structure with list of sync_job dicts
+class stopped_jobs_key_handle(json_redis_key_handle):
+    def __init__(self, redis_handle):
+        super().__init__(redis_handle, "irods_ingest_stopped_jobs", "")
```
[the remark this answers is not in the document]
alanking: Done

irods_capability_automated_ingest/sync_job.py
alanking: Added a mention of the JSON array in the docstring.

irods_capability_automated_ingest/sync_job.py
alanking: How's this look? Just added a `%z` to the end...
```javascript
    {
      "job_name": "acd6eb3a-84cc-11ef-8954-0242c0a84006",
      "total_tasks": 3,
      "remaining_tasks": 1,
      "failed_tasks": 0,
      "retried_tasks": 0,
      "elapsed_time": "00:00:33",
      "start_time": "2024-10-07 16:53:27+0000"
    }
```

MartinFlores751: That works. As a side note, if we like ISO 8601, I think we could sub the space with a `T` and have it be `2024-10-07T16:53:27+0000`.

This is completely optional, and imo more readable with the space.

korydraughn: This is the format of the timestamp written to the iRODS server log: `2024-10-06T16:34:50.259Z`

How do we feel about using the same format?

MartinFlores751: That's fine. We would have to put the 'Z' ourselves, and make sure the input is `gmtime(...)` otherwise the timezone could be off. Python seems to prefer the 'full offset' vs the 'Z' shorthand.

Maybe add a comment on the side to update the format string if we do change it in the future?

korydraughn: UTC and consistency are the main things to get right. I don't have strong feelings towards `Z` or `+0000`.

Deferring to @alanking.

alanking: The question I would have with respect to consistency is... with what?

Here's a table of all the things involved in this "stack", and they are all in different formats:
| Logger | date/time format |
| ---- | ---- |
| Postgres | `2024-10-07 18:35:22.539 UTC` |
| Redis | `07 Oct 2024 18:35:12.080` |
| Celery | `2024-10-07 17:27:57,490` |
| iRODS server | `2024-10-06T16:34:50.259Z` |
| Ingest log messages | `2024-10-07T17:27:57.490467+00:00` |
| sync_job info (this PR) | `2024-10-07 16:53:27+0000` |

I have no preference on what we end up using, but presently, no two timestamp formats are alike across all the logs. :(

korydraughn: > with what?

Timestamps throughout the ingest project. Normally, this isn't something we have to think about because there's only one logger involved.

As long as the iRODS specific code within the project uses the same format for timestamps, it's fine.

alanking: In that case, I suppose we ought to use the format in the ingest log messages, as seen here:
https://github.com/irods/irods_capability_automated_ingest/blob/0661effd516f79cfe08f414427cdba5097b88927/irods_capability_automated_ingest/sync_logging.py#L15-L19

It just defers to the ISO format as defined in the `datetime` library.

korydraughn: What does that produce?

alanking: Ingest log messages have a `@timestamp` field that looks like this: `2024-10-07T17:27:57.490467+00:00`

Here's what I get for this new thing when I use the `isoformat`:
```json
    {
      "job_name": "a911cb86-84e1-11ef-8e99-0242c0a86006",
      "total_tasks": 3,
      "remaining_tasks": 1,
      "failed_tasks": 0,
      "retried_tasks": 0,
      "elapsed_time": "0:11:28.102603",
      "start_time": "2024-10-07T19:23:41.089828+00:00"
    }
```
The `start_time` looks like the `@timestamp` fields in the ingest logs here. Note that this includes microseconds.

The `elapsed_time` contains microseconds because it is just printing out a `datetime.timedelta` which stores seconds and microseconds. We can format this to only include milliseconds, or just seconds, or whatever.

korydraughn: I think format used by the ingest logs is good. Just convert the microseconds to milliseconds and call it a day.

alanking: Done. Luckily, the `datetime.isoformat` function provides a `timespec` parameter to specify "milliseconds" precision, so this was quite simple. Also updated the log's formatting to reflect this.

korydraughn:
Anything left for this? Are the tests passing?

alanking:
I'll run the test suite once more for good measure, but these changes mainly affect the sync script rather than the framework, so these should probably be passing. I'll report back with results soon.

I would also like to update the commit message for "Reset Redis handles for stopped jobs", but that's about it for this one.

korydraughn:
Sounds good. Will wait for the test results before approving.

alanking:
Tests are passing.

alanking:
Squashed, tweaked commit message, and added a thing that Black formatter caught.

korydraughn (approved):
Pound it

alanking:
#'d, mergin

Issue #288, opened 2024-10-01
Add test for Delete Mode which deletes all files in a directory, but not the directory


alanking:
This will test to see whether the collection associated with the directory is removed or not. We expect it *not* to be removed because the directory was not removed.

Issue #280, opened 2024-09-27
Investigate need for `time.sleep` after calling `os.makedirs`


alanking:
When I was writing tests for #48, I ran across a strange situation in which the Celery workers - during a sync job - do not see a directory tree (with some files) that I created in a Python `unittest.setUp` method... unless I sleep for some amount of time between creating the directories and starting the sync job. Through some light experimentation, I have found that a sleep as little as 1 second keeps things working, but I do not know how short the sleep can actually be while things continue to work.

The directories being created are inside a Docker volume, so I'm not sure if that has anything to do with it.

alanking:
The same seems to be true for S3 buckets in some S3 sync tests I've been writing. Very strange.

trel:
sounds like systems are lying/optimizing and you're catching them in it...

alanking:
The issue is that the Celery workers are not actually ready to execute tasks in time when the tests start running. They take a full second or more to "mingle" (coordinate with other Celery nodes). It seems strange to me that this would cause problems (why not wait to do work until they're actually ready?) but we don't need to mingle in the test workers anyway because they assume they are "all alone".

Issue #100, opened 2019-09-03
error adding metadata to data object with grave character in the path


trel:
Using v0.3.7.

When scanning an NFS-mounted directory, and gathering metadata for every file found...

Full error log is here:
```json
{
  "task_id": "9994fa13-ce42-11e9-b20c-0aaaf8493372",
  "task": "sync_dir",
  "einfo": "<ExceptionInfo: CAT_UNKNOWN_FILE(None,)>",
  "path": "/full/path/to/dirwith`/subdir",
  "exc": "CAT_UNKNOWN_FILE(None,)",
  "traceback": [
    [
      "/home/irods/venv/lib/python3.4/site-packages/celery/app/trace.py",
      382,
      "trace_task",
      "R = retval = fun(*args, **kwargs)"
    ],
    [
      "/home/irods/venv/lib/python3.4/site-packages/celery/app/trace.py",
      641,
      "__protected_call__",
      "return self.run(*args, **kwargs)"
    ],
    [
      "/home/irods/venv/lib/python3.4/site-packages/irods_capability_automated_ingest/sync_task.py",
      388,
      "sync_files",
      "sync_entry(self, meta, \"file\", sync_irods.sync_data_from_file, sync_irods.sync_metadata_from_file)"
    ],
    [
      "/home/irods/venv/lib/python3.4/site-packages/irods_capability_automated_ingest/sync_task.py",
      482,
      "sync_entry",
      "raise self.retry(max_retries=max_retries, exc=err, countdown=retry_countdown)"
    ],
    [
      "/home/irods/venv/lib/python3.4/site-packages/celery/app/task.py",
      667,
      "retry",
      "raise_with_context(exc)"
    ],
    [
      "/home/irods/venv/lib/python3.4/site-packages/irods_capability_automated_ingest/sync_task.py",
      474,
      "sync_entry",
      "datafunc(meta2, logger, True)"
    ],
    [
      "/home/irods/venv/lib/python3.4/site-packages/irods_capability_automated_ingest/sync_irods.py",
      415,
      "sync_data_from_file",
      "call(hdlr_mod, \"on_data_obj_create\", register_file, logger, hdlr_mod, logger, session, meta, **options)"
    ],
    [
      "/home/irods/venv/lib/python3.4/site-packages/irods_capability_automated_ingest/sync_utils.py",
      147,
      "call",
      "getattr(hdlr_mod, hdlr)(func, *args, **options)"
    ],
    [
[...]

Issue #282, opened 2024-09-27
Add Delete Mode for S3 buckets


alanking:
#48 adds Delete Mode for filesystem syncs. This issue is to address adding Delete Mode support for S3 bucket syncs.

PR #300, opened 2024-10-08
Adding and fixing some tests

alanking:
Addresses #100 
Addresses #280 
Addresses #288 

All tests are passing.

### DIFF
--- a/irods_capability_automated_ingest/test/test_delete_modes.py
+++ b/irods_capability_automated_ingest/test/test_delete_modes.py
@@ -70,6 +70,37 @@ def wait_for_job_to_finish(workers, job_name, timeout=60):
     )
 
 
+def get_event_handler(operation, delete_mode):
+    operation_strings = {
+        Operation.NO_OP: "NO_OP",
+        Operation.REGISTER_SYNC: "REGISTER_SYNC",
+        Operation.REGISTER_AS_REPLICA_SYNC: "REGISTER_AS_REPLICA_SYNC",
+        Operation.PUT: "PUT",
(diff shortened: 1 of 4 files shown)


alanking:
Added a bail-out for use of DeleteMode with S3 bucket syncs. #282 will add actual support.

irods_capability_automated_ingest/test/test_delete_modes.py
```diff
@@ -592,6 +577,56 @@ def test_PUT_APPEND_and_NO_TRASH_deletes_collections(self):
...
+                self.assertFalse(
+                    self.irods_session.data_objects.exists(
+                        "/".join([self.target_subcollection_for_removal, file])
+                    ),
+                )
+            # The collection should still exist.
+            self.assertTrue(
+                self.irods_session.collections.exists(
+                    self.target_subcollection_for_removal
+                ),
```
korydraughn, line 627: So the black formatter is adding commas to the end of functions too?
Does `assertTrue` accept a second parameter (i.e. a message or something)?

korydraughn (approved):
Changes look good.

irods_capability_automated_ingest/test/test_delete_modes.py
alanking: Yes to both. Black formatter is doing most of these line separations and comma-additions. `assertTrue` accepts an optional `msg` parameter: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue

alanking:
#'d, mergin.

Issue #261, opened 2024-09-11
Integrate character_map / UnicodeEncodeError paths with delete operation feature


alanking:
Requires https://github.com/irods/irods_capability_automated_ingest/issues/48 to be completed first.

trel:
Hmm, can `character_map` and `delete` be detected and declared incompatible?

That doesn't really help with the `UnicodeDecodeError` cases...

alanking:
Right. The options I discussed with @korydraughn were...

1. Just implement the support (#261, this issue). This will take some effort.
2. Detect and prevent. This is easy enough for `character_map`. From what I can tell, the amount of effort to do this for the `UnicodeEncodeError` case would be high enough that we might as well just implement the support.
3. Punt. Delete Mode is a new feature and we are still pre-1.0, so "Use at your own risk."

We wanted 2 to be easy, but I don't think it will be much less effort than 1. 3 is indefensible given the high risk of data loss.

trel:
That assessment seems correct to me.

korydraughn:
Let's give it a try and implement option 1.

alanking:
After I started implementing this, I found that option 2 was actually quite easy. We are just going to do that and kick this back to 0.7.0. Sorry for turbulence.

PR #302, opened 2024-10-10
[#261] Prevent Delete Mode with mapped paths 

alanking:
In service of #261

### DIFF
--- a/irods_capability_automated_ingest/tasks/filesystem_tasks.py
+++ b/irods_capability_automated_ingest/tasks/filesystem_tasks.py
@@ -217,17 +217,36 @@ def filesystem_sync_path(self, meta):
 
         delete_extraneous_items = utils.DeleteMode.DO_NOT_DELETE != delete_mode
         if delete_extraneous_items:
-            subcollections_in_collection, data_objects_in_collection = (
-                get_collections_and_data_objects_in_collection(
-                    meta, destination_collection
+            # TODO(#261): Prevent Delete Mode usage with character_map until support is implemented.
+            character_map = getattr(event_handler.get_module(), "character_map", None)
+            if character_map:
+                logger.warning(
+                    "character_map and non-default Delete Mode are incompatible at this time. "
+                    "Delete Mode will be disabled for this task."
                 )
-            )
+                delete_extraneous_items = False
+            else:
+                subcollections_in_collection, data_objects_in_collection = (
+                    get_collections_and_data_objects_in_collection(
+                        meta, destination_collection
+                    )
+                )
         else:
             subcollections_in_collection = []
             data_objects_in_collection = []
 
         for obj in itr:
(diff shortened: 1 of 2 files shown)


irods_capability_automated_ingest/tasks/filesystem_tasks.py
```diff
@@ -217,18 +217,37 @@ def filesystem_sync_path(self, meta):
...
 
         for obj in itr:
             full_path = os.path.abspath(obj.path)
 
+            # TODO(#261): Prevent Delete Mode usage when any path is modified due to UnicodeEncodeError until support is
+            # implemented.
+            if delete_extraneous_items:
+                if utils.is_unicode_encode_error_path(full_path):
+                    logger.warning(
+                        "Found a path which will be modified due to UnicodeEncodeError. "
```
alanking, line 246: We cannot print the path in the error message because if we could it wouldn't raise a `UnicodeEncodeError` :)

korydraughn (commented):
Looks good.

Are we ready for pounds?

alanking:
Yes, I think so. Confirmed that Delete Mode tests are passing, which is all this should effect.

irods_capability_automated_ingest/tasks/filesystem_tasks.py
trel: could... print the created/safe version?   not a blocker.

korydraughn (approved):
Pound it.

irods_capability_automated_ingest/tasks/filesystem_tasks.py
alanking: If we translate the path, we might as well go ahead and implement the feature because we've done the work... right?

This path mapping is not modularized:
https://github.com/irods/irods_capability_automated_ingest/blob/972cdefc2e927eeb015695c5a7b5cd66bac84bdf/irods_capability_automated_ingest/tasks/filesystem_tasks.py#L376-L380

alanking: If a path is translated due to `UnicodeEncodeError`, a warning message already appears which alerts the user to the fact that the translation occurred and what the new path is:
https://github.com/irods/irods_capability_automated_ingest/blob/972cdefc2e927eeb015695c5a7b5cd66bac84bdf/irods_capability_automated_ingest/tasks/filesystem_tasks.py#L387-L390

I think for the message about Delete Mode, it doesn't really matter which path initially disabled Delete Mode. The user can find out which paths were translated via the aforementioned warning message.

All that being said, we can just copy the call to `str.encode` if desired. Up to you.

trel: nah, leave it.   just get it into the 'real' issue where we do the work for later.

alanking:
#'d. Mergin

Issue #180, opened 2022-09-30
Housekeeping - documentation, typos, configuration


d-w-moore:
And perhaps other miscellaneous, minor and cosmetic things.

PR #308, opened 2024-10-28
[#180] Update requirements.txt

alanking:
This file has not been updated in a while, as you can see. Perhaps we need to start doing this before each release to make sure everything is the right version.


trel:
do we need to declare all these things?

can/shouldn't we leave the dependencies of dependencies out?   are there any of those?

alanking:
The setup.py file is the thing driving dependencies. 0.6.0 has the following explicit dependencies:
https://github.com/irods/irods_capability_automated_ingest/blob/b96fc67b824e050840eb8f47720bde922b06b936/setup.py#L43-L50

0.6.0 actually uses `os.scandir`, so we don't even need that `scandir` dependency. The rest of these are used directly by the ingest tool.

Everywhere I've seen on the Internet says that this is how you generate a requirements.txt for a project:
```
pip freeze > requirements.txt
```
So, this is what I've done here with a clean environment which simply installed the `irods-capability-automated-ingest` pip package. The `pip freeze` documentation itself employs this practice as well: https://pip.pypa.io/en/stable/cli/pip_freeze/#examples

Willing to try only specifying the direct dependencies, but that does not appear to be standard practice.

trel:
i see - that does make sense.       yes, please try a clean install without `scandir` and... then yes, i think we're good.

thanks.

alanking:
Removed `scandir` dependency from `setup.py`, ran `pip freeze > requirements.txt` again, and tests are running now.

alanking:
Tests passed.

trel:
looks good to me - i think it's ready.

alanking:
Okey dokey. Number go up

Issue #310, opened 2025-11-19
Apply Ruff to repo


alanking:
Let's go ahead and apply Ruff, as we are thinking about doing to python-irodsclient (see https://github.com/irods/python-irodsclient/issues/726). This repo has already been formatted using Black: 94b0efb478d922b5e1e65144d269904f795834a7 The Ruff formatter is based on Black, so very little change will occur.

If we want to change the formatting or the linter or what have you, we can argue about those configurations. Ruff offers a pretty vast array of linter / formatter Rules (which includes those from flake8 and some other linters): https://docs.astral.sh/ruff/rules I'm sure we can come up with some combination that makes everyone happy.

PR #311, opened 2025-11-19
[#310] Run Ruff check and format on codebase

alanking:
Addresses #310

Test suite passed.

Please have a look at the config file and see whether there are any tweaks we want to make. The one shown is the default, but we can modify it and point the workflows at it.

---

Some thoughts:

I didn't do very much work to apply these fixes. Ruff did most of it itself. The GitHub Actions workflow uses `--diff`, so it will not modify anything in PRs. I might update the README to include a tidbit about running Ruff locally because that was incredibly easy as well. I just ran the docker container as described here: https://docs.astral.sh/ruff/installation

Here's what running the check (linter) looks like:
```
docker run -v .:/io --rm ghcr.io/astral-sh/ruff check --fix
```
I included `--unsafe-fixes` as well to try to fix everything and then combed through the changes and fixed what didn't quite work out the way it should have. You can of course just use `--diff` to only display a diff like in GitHub Actions.

And here's what running the formatter looks like:
```
docker run -v .:/io --rm ghcr.io/astral-sh/ruff format
```

I consider this a forerunner to using Ruff in the python-irodsclient (https://github.com/irods/python-irodsclient) repo. We can try it out here to see if we like it, and, if so, we can apply it there as well.

### DIFF
--- a/.github/workflows/ruff-action.yaml
+++ b/.github/workflows/ruff-action.yaml
@@ -1,0 +1,26 @@
+name: ruff-action
+
+on: pull_request
+
+jobs:
+  ruff-format:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+    
+      - name: Run Ruff formatter (check only)
+        uses: astral-sh/ruff-action@v3
+        with:
+          args: "format --check --diff"
+
+  ruff-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Run Ruff check (check only)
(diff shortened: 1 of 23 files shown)


alanking:
You can see the GitHub Actions workflow results over in a test PR I made in my fork: https://github.com/alanking/irods_capability_automated_ingest/pull/1 Not sure why they won't run here - guess it's different because it's not my repo.

EDIT: And here's what it looks like if the problems aren't gone: https://github.com/alanking/irods_capability_automated_ingest/pull/2

alanking:
Converting to draft until https://github.com/irods/python-irodsclient/issues/726 is a bit further along. We will want to do something similar if not identical in this repo.

Issue #313, opened 2025-12-08
urllib3 2.6.0 causes break in S3 scanner


alanking:
urllib3 interface was changed in 2.6.0: https://urllib3.readthedocs.io/en/stable/changelog.html#removals

We use this method here:
https://github.com/irods/irods_capability_automated_ingest/blob/b1c236fbb8bd39e99ada6775a693719cf6e02a5e/irods_capability_automated_ingest/irods/s3_bucket.py#L172

We do not have any restrictions on urllib3's version right now, so we need to fix this.

PR #314, opened 2025-12-08
[#313] Fix urllib3.response.HTTPResponse interface change

alanking:
Addresses #313

Required before merging #312

All tests pass with this change (for urllib3==2.5.0 and urllib3==2.6.0). I don't know if we should be in the business of attempting to span multiple versions of our dependencies, but that's what I'm doing here.


korydraughn (approved):
Looks fine to me.

Pound it if it's working.

alanking:
#'d, mergin

Issue #318, opened 2025-12-08
Add GitHub Actions Workflow for running tests


alanking:
This should run for at least every PR.

Issue #316, opened 2025-12-08
Support latest python-irodsclient


alanking:
The current PRC version must be less than 3.0.0:
https://github.com/irods/irods_capability_automated_ingest/blob/dccba7a55c396da6fc02381c250093684df07137/setup.py#L44

In order to get all the good things that have come out since then and support iRODS 5, we need to update this.

PR #319, opened 2025-12-08
Add GitHub Actions Workflow for running tests, update PRC

alanking:
Addresses #316
Addresses #318

Tried to keep the tests mostly in tact. We can of course improve these over time.

I created https://github.com/irods/irods_reusable_github_workflows/issues/15 to address the repeated copying and pasting of the irods-catalog-provider service in all of our projects.


alanking:
Here's what the test runs look like: https://github.com/alanking/irods_capability_automated_ingest/actions/runs/20040915625

docker/ingest-test/README.md
```diff
@@ -21,14 +21,15 @@ Note that, depending on whether the iRODS major/minor version is 4.2 or 4.3, the
...
-The `--env-file` option is required in order to correctly configure the environment for the tests.
+The test suite is one of the services of the docker-compose project, so it will run on its own. The container is tied to the tests running, so it will exit once completed. The `--env-file` option is required in order to correctly configure the environment for the tests.
 
-## Step 3: Bring down the project
+If you wish to leave the other services running after the test run, omit the `--abort-on-container-exit` option in the command above.
 
-The project is not made to come down by itself (yet), so it has to be brought down after each run:
+## Step 3: Cleaning up after tests finish
+
+The tests expect to have clean environment each time they are run. As such, we need to make sure to clean up after each test run. This can be done like this:
```
korydraughn:
```suggestion
The tests expect to have a clean environment each time they are run. As such, we need to make sure to clean up after each test run. This can be done like this:
```

docker/ingest-test/README.md
```diff
@@ -21,14 +21,15 @@ Note that, depending on whether the iRODS major/minor version is 4.2 or 4.3, the
 
 Bring up the docker-compose project and the test suite will run on its own:
 ```
-docker compose --env-file icommands.env up
+docker compose --env-file icommands.env up --abort-on-container-exit
```
korydraughn, line 24: So the additional option instructs docker compose to shut down when any of the containers terminate?

I assume that's a convenience thing?

docker/ingest-test/irods_catalog_provider/Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM ubuntu:22.04
```
korydraughn: Could bump to ub24?

korydraughn (commented):
Looks correct.

setup.py
```diff
@@ -41,7 +41,7 @@
     packages=find_packages(),
     install_requires=[
         "minio",
-        "python-irodsclient<3.0.0",
+        "python-irodsclient<4.0.0",
```
trel, line 44: should it be > 3.0.0 now?

or just... without version information?

docker/ingest-test/README.md
alanking: For manual / bench test runs, yes, this is just for convenience.

For the GitHub Actions Workflow, it's actually required. The test image runs the tests as its entrypoint while the rest of the services just stay up forever. So this signals to the other services that they need to end once the tests end. That allows the workflow to move to the `docker compose down` step.

docker/ingest-test/README.md
alanking: Done

docker/ingest-test/irods_catalog_provider/Dockerfile
alanking: I didn't realize that we released 4.3.4 on Ubuntu 24.04. I'll try bumping it.

setup.py
alanking: I don't think so. Everything except the tests will continue to work with <3.0.0, and even that is just a matter of a difference in Python types.

If we want to take off the version requirements, that's fine, but I think I'd prefer to lock it to less than the next major version so that we don't have sudden interface breaks.

alanking: Maybe @d-w-moore has an opinion on this matter

trel: i defer - just caught my eye.

korydraughn:
Noting here that the goals of this work are:
- Confirm the project works with declared supported versions of Python
- Confirm the project is always in a state where the tests pass

@alanking Are there other goals which aren't listed here?

Now for a question ... Do we need to test this against multiple versions of iRODS if PRC testing already covers that?

docker/ingest-test/irods_catalog_provider/Dockerfile
alanking: Confirmed that the Ubuntu 24.04 image builds. Will watch to see if tests pass here: https://github.com/alanking/irods_capability_automated_ingest/actions/runs/20043520393

docker/ingest-test/README.md
korydraughn: Cool cool.

setup.py
korydraughn: Adding my two cents ... I agree that applying some version info is good. It at least helps us have some idea of what is being tested.

alanking:
> Noting here that the goals of this work are:
> 
>     * Confirm the project works with declared supported versions of Python
> 
>     * Confirm the project is always in a state where the tests pass
> 
> 
> @alanking Are there other goals which aren't listed here?

For this PR, that covers what I was trying to do, I think. I don't know that I fully understand what you mean by the second bullet point, but I took it to mean having automated testing in PRs so that we can merge changes more quickly (i.e. dependabot PRs).

> Now for a question ... Do we need to test this against multiple versions of iRODS if PRC testing already covers that?

I keep going back and forth on this. I think we should at least strive to test against iRODS 5, whether that's in addition to or instead of the 4.3.x testing we have right now.

korydraughn:
The second bullet is about always being close to a release - i.e. our GHAs should always be green if we're paying attention and fixing bugs.

As for testing against iRODS 5, it doesn't hurt to do that. I can see it going either way. And seeing the tests pass against iRODS 5 can only raise confidence in the software.

docker/ingest-test/irods_catalog_provider/Dockerfile
alanking: Tests passed. Resolving.

alanking:
> The second bullet is about always being close to a release - i.e. our GHAs should always be green if we're paying attention and fixing bugs.

Understood. In that case, should we be running this nightly or weekly, or something? Unless a push or PR happens, we won't know that the tests are passing until we run them. Python dependencies, unfortunately, move around a lot, unless you lock them down tight. And then it becomes a nightmare to manage.

> As for testing against iRODS 5, it doesn't hurt to do that. I can see it going either way. And seeing the tests pass against iRODS 5 can only raise confidence in the software.

Okay. I created this issue to deal with testing iRODS 5: https://github.com/irods/irods_capability_automated_ingest/issues/320

korydraughn:
> Understood. In that case, should we be running this nightly or weekly, or something? Unless a push or PR happens, we won't know that the tests are passing until we run them. Python dependencies, unfortunately, move around a lot, unless you lock them down tight. And then it becomes a nightmare to manage.

True. I don't know that we need to run the tests on a nightly basis, but a few times a week seems okay. I take it you're referring to released versions of iRODS for nightly/weekly test runs?

trel:
we can also have the button to run-on-demand.

korydraughn:
That's correct.

https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow

alanking:
I have successfully implemented a `workflow_dispatch` (run-on-demand). I'm playing around with `schedule`, but yet to see it work. We can always include it to run weekly and check in occasionally to see if it ever runs automatically. :) Should I include those changes in this PR?

alanking:
After some more fiddling, I did see a scheduled job run. So, I've included the changes here.

setup.py
trel: okay, so resolving... good as is...

trel (approved):
squash em.

alanking:
Squashed

korydraughn (approved):

korydraughn:
Pound it.

alanking:
#'d, mergin

